PR #7, opened 2022-07-22
The use of class_alias is not a side effect

andrewnicols:
A file containing a class_alias should not be considered a side-effect
on its own.

We may want to use these to migrate away from use of lib.php in the
longer term.

Covered with unit tests.


stronk7 (approved):
Thanks Andrew, this makes sense. Approved.

PR #31, opened 2022-09-27
Use Squizlabs Commenting sniffs for Method PHPDoc

### LINKED ISSUE #30: Migrate moodle-local_moodlecheck phpdoc checks to this standard
I think we can probably migrate most of our custom-written phpdoc checks into this standard.

There are several, but (un)helpfully phpcs sucks at documenting the rules available. Here's a summary of some of them:
https://stackoverflow.com/questions/13767550/is-there-a-phpcs-standard-targeting-php-docblocks

We will need to tailor this a little, but it should be able to capture most of the phpdoc sniffs for methods at the very least.

andrewnicols:
This commit addss support for the standard Commenting sniffs relating to methods only.

Some of the specific sniffs have been disabled, or their severity/type changed to suit our rules.

Fixes #30


stronk7:
Maybe we can discard this now that we have #30 to make it in an ordered and planned way?

andrewnicols:
Agreed.

I've also found that most of those sniffs actually are not suited. They do what we want, and then some other things too (grr)

PR #32, opened 2022-10-13
Bump to actions/checkout@v3

stronk7:
Related to this deprecation of NodeJS 12 for GHA actions:

https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

While it doesn't affect to this repository, let's bump to the new v3 version now so we keep it updated.

### DIFF
--- a/.github/workflows/phpcs.yml
+++ b/.github/workflows/phpcs.yml
@@ -15,7 +15,7 @@ jobs:
           - php: 7.4
     steps:
       - name: Check out repository code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Setup PHP ${{ matrix.php }}
         uses: shivammathur/setup-php@v2
--- a/.github/workflows/phpcs.yml
+++ b/.github/workflows/phpcs.yml
@@ -24,7 +24,7 @@ jobs:
           extensions: ${{ matrix.extensions }}
           ini-values: pcov.directory=moodle
           coverage: pcov
-          tools: composer, phpcpd
+          tools: composer
 
       - name: Install composer dependencies
         run: |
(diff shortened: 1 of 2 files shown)


stronk7:
Note that the existing warnings @ https://github.com/moodlehq/moodle-cs/actions/runs/3245878158 about `The `set-output` command is deprecated...` are ignorable. The fix has been already applied to the php-setup action and it's going to be released this weekend, so next week that warning will be gone.

What I don't understand are the exit codes in that run, how did it pass but exited with error? In any case that's 100% unrelated with this PR.

Ciao :-)

jrchamp:
Do you mean the `continue-on-error: true`?

stronk7:
> Do you mean the `continue-on-error: true`?

I mean, it's quite annoying to see this:

green_and_red

Looking within the runs, I think it's the copy paste job the one failing and leading to that. IMO if they fail we should fail, else it's futile to have that check.

I'm going to see if can avoid the detection within fixtures and then, enable real failures to happen...

stronk7:
I've added a 2nd commit to the candidate branch that moves the installation of phpcpd as a new dev requirement (so people can also install it locally in an easy way), fixed the exclusion to be "moodle/Tests" (we recently moved to it but missed that case) and also, make the check to fail on error, so we detect problems as they happen.

Surely we'll need some local script (make or whatever) to be able to run all the CI checks locally, but that's out from this issue scope.

What do you think about the changes, @andrewnicols ?

stronk7:
I'm going to proceed with this (agreed with @andrewnicols , thanks!). Need it landed for some related experiments.

Ciao :-)

andrewnicols (approved):

PR #36, opened 2023-01-12
Integration test & PHPCompatibility bump to develop#2fb82334

stronk7:
Current PHPCompatibility 9.3.5 is old (2019) and it's not PHP 8.1 compliant (neither has many specific sniffs for new versions).

No matter the lack of new releases, the development is active and happening under their "develop" branch. And it supports PHP 8.1, 8.2 and has a good number of Sniffs for those versions.

When trying to run current moodle-cs from other tools (moodle-plugin-ci, from core...) with PHP 8.1 we were getting some deprecated warnings leading to failed (phpcs and phpcbf) executions.

So, here we are:

1. Updating to PHPCompatibility develop#2fb82334
2. Adding some simple integration tests: Ensure that the GH workflow always run the commands (phpcs and phpcbf)) and they work as expected.
3. Add back the `Squiz.Arrays.ArrayBracketSpacing` sniff to the moodle standard. Somehow it was forgotten/missed when moving the standard from codechecker to moodle-cs.

You can see the integration tests failing [here](https://github.com/stronk7/moodle-cs/actions/runs/3904744318), that is this branch with the PHPCompatibility upgrade disabled. Once the upgrade is applied, all tests are passing.

Ciao :-)

### DIFF
--- a/.github/workflows/phpcs.yml
+++ b/.github/workflows/phpcs.yml
@@ -22,7 +22,7 @@ jobs:
         with:
           php-version: ${{ matrix.php }}
           extensions: ${{ matrix.extensions }}
-          ini-values: pcov.directory=moodle
+          ini-values: pcov.directory=moodle, error_reporting=-1, display_errors=On
           coverage: pcov
           tools: composer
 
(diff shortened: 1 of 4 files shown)


andrewnicols (approved):

stronk7:
Yay, thanks!

Issue #1, opened 2022-04-25
Mockup a mini-moodle-site to get some part of the Sniffs covered


stronk7:
While the Sniffs (and their tests) can run without Moodle, some of them run better with a Moodle checkout at hand. That way they are able to detect branches, components and other details.

But checking out Moodle is really slow and expensive.

So, the idea is to be able to create a mini-moodle-site, having all the requirements (mocked version.php, config-dist, core_component...) . So the testing depending on those files can happen normally.

In fact, surely we should be able to have different mini-moodle-sites, mocking different branches, because some Sniffs change their behavior based on that.

Note this is not critical (it can also be checked @ locacl_codechecker, where a complete moodle-site is available), but better do it here (self-contained within moodle-cs).

Ciao :-)

stronk7:
To say it in another way... just compare the code-coverage current reports:
- here
- @ local_codechecker

And we need to achieve the same coverage here (no skipped tests).

stronk7:
Note this can be closed once #49 is done.

Ciao :-)

PR #49, opened 2023-05-18
Enable skipped tests

stronk7:
This enables the MoodleUtil tests that were skipped since we moved from local_codechecker to standalone moodle-cs repository.

Basically it uses both streamvfs and a mini-mocked site (with the minimum files required) to be able to ensure that the MoodleUtil class is able to find the moodle root directory, the current moodle branch being used and the component that belongs to every file (by using core_component).

Also, some old code (only needed when this was part of a standard plugin) has been removed, not needed anymore.

Finally, some variables in the tests have been renamed to avoid future confusions and all this should lead to a very good raise in phpunit coverage.

This is needed to start creating more methods in the MoodleUtil class, able to load valid apis that, later, will be used by all the namespace sniffs.

Ciao :-)


stronk7:
Fun, this closes #1 , yay!

stronk7:
Added one more commits that adds testing to 4 old Sniffs that were not covered at all. That makes coverage to sky-rocket to 95%, yay!

andrewnicols (approved):

stronk7:
tata!

PR #54, opened 2023-08-28
Update composer dependencies to actual ones

stronk7:
- All them must work for php74 and up.
- PHPCompatibility bump to current dev version: 0a17f9ed

Also, update the CHANGELOG file.


stronk7:
Self-merged in order to make a release today.

PR #56, opened 2023-09-14
Rename MoodleCSBaseTest to MoodleCSBaseTestCase

stronk7:
As a safety check, abstract classes are not allowed to end with [Tt]est anymore in PHPUnit 10. We were doing that with the `MoodleCSBaseTest` class, hence moving to `TestCase` suffix.

Also, the `MoodleCSStandardTest` was incorrectly named (not matching the file name that was correct), so
renamed it to `MoodleStandardTest`.

These changes come onboard before start adding APIs support to the standard (first in PHPUnit classes and, once working) to the whole codebase (that's the plan).


paulholden (approved):
LGTM

Issue #45, opened 2023-03-14
Strict Standard option


jrchamp:
#44 has highlighted some of the challenges in addressing both a large, preexisting codebase and the inconsistency issues that come from loose rules for developers. The current one-at-a-time policy queue is working to slowly address these changes publicly, but it would be nice to have a strict standard option now that complies with Moodle's Coding style and the non-conflicting elements of the PSR12 and PSR1 standards.

Having a strict standard would allow us to get more consistent results from both contributors and vendors.

Implementation would likely involve:

- Creating a moodle-strict standard in this repository (it could inherit from the moodle standard and apply stricter rules)
- Adding a `--strict` or similar flag to moodle-plugin-ci so that the moodle-strict standard can be used

I already have something close, so I'm more wondering if this is something you would consider for inclusion.

stronk7:
Hi, I'm closing this as fixed by #57

Issue #25, opened 2021-01-15
[CONTRIB-8097] Enforce coding guidelines for trailing comma in multi-line array declaration


sarjona:
This is about to enforce the following policy decided recently (https://tracker.moodle.org/browse/MDLSITE-5873).

> Policy: Trailing comma in multi-line array declaration
> Since 20th May 2020, this rule applies:
> 
> 1. The uses of a trailing comma on multi-line array declaration is required in PHP and JS.
> 2. It's NOT required to change all current old-style occurrences in the codebase. We'll follow a progressive approach here (new code and modified code)
> For example for the expected format:
> 
> $foo = [
>     'a',
>     'b',
>     'c',
> ];
>

This issue is about to verify these details in the codechecker. Nothing more, nothing less.

---
Reporter: Sara Arjona
Original issue: https://tracker.moodle.org/browse/CONTRIB-8097

andrewnicols:
MediaWiki have a rule we can use:

https://packagist.org/packages/mediawiki/mediawiki-codesniffer

```
+    <rule ref="MediaWiki.Arrays.TrailingComma">
+      <properties>
+        <property name="singleLine" value="false" />
+        <property name="multiLine" value="true" />
+      </properties>
+    </rule>
```

jrchamp:
https://github.com/PHPCSStandards/PHPCSExtra has a significant number of sniffs, including one for this: NormalizedArrays.Arrays.CommaAfterLast

stronk7:
While I'm not a fan of adding too many standards from outside (as commented in #34), the [PHPCSExtra](https://github.com/PHPCSStandards/PHPCSExtra) is really interesting, specifically the `Universal` rules, easy to add to our ruleset, but also because it comes from Juliette, also super involved in the PHPCompatibility one, also the composer plugin that we use to install it and other things...so somebody to trust, I'd say.

Also, somehow related... I've been thinking about #45 and, maybe, the way to go is something similar to the schema used by `PHPCSExtra`, offering different rulesets (standards) and allow people to adhere to them (say current "moodle", the general one, then a "moodle-strict" or a "moodle-crazy/next" or so...).

Ciao :-)

andrewnicols:
I'ma big fan of that approach. I have a local schema `moodle-strict` which does some of this but I'd prefer someting in core.

stronk7:
BTW,

[...]

PR #57, opened 2023-09-14
Add a new moodle-extra coding style

andrewnicols:
As dicussed in #45, and #25 it would be good to have a strict option.

This is an attempt largely based on the one I use daily (https://github.com/andrewnicols/moodle-cs-strict) with a few additions and removals.


stronk7:
I like this, and we can go promoting things from the "extra" ruleset to the "normal" one while things evolve...

BTW, does "extra" sound ok for you? For some reason (not sure why, to be honest, maybe because of Spanish...) I'm not sure. Maybe "plus", "strict", "get_it_serious" (lol)...

In any case I'm not strong at all with the above, if you're happy with "extra", I'm happy too. Just tell and I will proceed.

Ciao :-)

andrewnicols:
> BTW, does "extra" sound ok for you? For some reason (not sure why, to be honest, maybe because of Spanish...) I'm not sure. Maybe "plus", "strict", "get_it_serious" (lol)...

I went for `extra` because that's what WordPress do for theirs: https://github.com/WordPress/WordPress-Coding-Standards#standards-subsets

Their description is:
> extended ruleset with recommended best practices, not sufficiently covered in the WordPress core coding standards

I've also taken the liberty of updating the readme here to:
- describe both of the coding tsyles
- update information on how to use the new coding styles
- remove the old information on configuration (moodle-local_codecheck is now wrong, and the other options are far from ideal).

stronk7:
Here we go! (note, I'll be making a release later today if that's ok for you)

andrewnicols:
Works for me.

I considered adding some more rules but I feel we should add unit tests for what we have first.

I was also wondering about writing a sniff for data providers to make them statically defined.

stronk7 (approved):
Thanks @andrewnicols,

it looks great!

PR #60, opened 2023-09-17
Require trailing commas in multi-line Arrays

### LINKED ISSUE #25: [CONTRIB-8097] Enforce coding guidelines for trailing comma in multi-line array declaration

andrewnicols:
Fixes #25

This was agreed in May 2020, so all branches since Moodle 3.9 should use this. We can now safely make this mainstream and not just in moodle-extra.

### DIFF
--- a/moodle-extra/ruleset.xml
+++ b/moodle-extra/ruleset.xml
@@ -48,9 +48,6 @@
     <!-- Detect duplicate array keys -->
     <rule ref="Universal.Arrays.DuplicateArrayKey"/>
 
-    <!-- Require a comma after the last element in a multi-line array, but prevent in a single-line array definition -->
-    <rule ref="NormalizedArrays.Arrays.CommaAfterLast"/>
-
     <!-- Disallow use of list() instead of [] -->
     <rule ref="Universal.Lists.DisallowLongListSyntax"/>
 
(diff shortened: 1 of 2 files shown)


stronk7 (approved):

timhunt:
Note, the sniff we have chosen does not implement then Moodle policy.

Arrays like

```php
        $crossword = new backup_nested_element('crossword', ['id'], ['correctfeedback',
            'correctfeedbackformat', 'numrows', 'numcolumns', 'accentgradingtype', 'accentpenalty',
            'partiallycorrectfeedback', 'partiallycorrectfeedbackformat',
            'incorrectfeedback', 'incorrectfeedbackformat', 'shownumcorrect']);
```

have always been considered fine, but the sniff complains about them.

PR #61, opened 2023-09-17
Check namespaces and class imports for leading backslash

### LINKED ISSUE #20: [CONTRIB-7828] Implement checks for namespace/import statements
Would be useful if the [guidelines regarding namespaces ](https://docs.moodle.org/dev/Coding_style#Namespaces)were enforced with Code checker, for example:

> Never use a leading backslash (\) in "namespace" and "use" statements.

---
Reporter: Paul Holden
Original issue: https://tracker.moodle.org/browse/CONTRIB-7828

andrewnicols:
Fixes #20

### DIFF
--- a/.phplint.yml
+++ b/.phplint.yml
@@ -2,3 +2,4 @@ path: ./
 exclude:
   - vendor
   - moodle/Tests/fixtures
+  - moodle/Tests/Sniffs/Namespaces/fixtures
--- a/moodle/Sniffs/Namespaces/NamespaceStatementSniff.php
+++ b/moodle/Sniffs/Namespaces/NamespaceStatementSniff.php
@@ -1,0 +1,71 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Checks that each file contains the standard GPL comment.
+ *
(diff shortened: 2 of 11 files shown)


stronk7:
Q: Shouldn't we also add `PSR12.Files.ImportStatement.LeadingSlash` to get imports also covered?

And little detail, it looks "strange" that our class is named "NoLeadingSlash", shouldn't it be better named "LeadingSlash". 

Similar sentiment with the problem name, we are naming it "NamespaceStartsWithSlash", that is redundant ("namespace") and uses different wording. Why not, also for the problem, just use "LeadingSlash".

FYC, ciao :-)

PS, I've seen that there are a few tests failing right now...

andrewnicols:
> Q: Shouldn't we also add PSR12.Files.ImportStatement.LeadingSlash to get imports also covered?
Bah - I originally applied this to the extra ruleset which already has it. Updated to include it/.

> And little detail, it looks "strange" that our class is named "NoLeadingSlash", shouldn't it be better named "LeadingSlash".
> 
> Similar sentiment with the problem name, we are naming it "NamespaceStartsWithSlash", that is redundant ("namespace") and uses different wording. Why not, also for the problem, just use "LeadingSlash".

If I make this LeadingSlash and name the problem `LeadingSlash` it becomes `moodle.Namespace.LeadingSlash.LeadingSlash` which is also redundant.

Rather than that, I've tried to take inspiration from the PSR-12 one and now have:
`moodle.Namespace.NamespaceStatement.LeadingSlash`

I think the duplication of `Namespace` in both the Sniff type, and the sniff name is okay because it's a check of the Namespace statement and naming it simply `Statement` could be confusing - `moodle.Namespace.Statement` is fine, but `Statement` on its own may confuse.

Thoughts?

> PS, I've seen that there are a few tests failing right now...

phplint was actually picking this up already so I think I've got that now!

stronk7 (approved):
I like making it `NamespaceStatement` so we can add later more checks there.

Note: I'm completing the API checks for PHPUnit classes (that already have Namespace checks). Once working I plan to move them from being "phpunit only" to be "whole codebase" ones. Surely then we'll be adding more checks to the `NamespaceStatement` sniff added here.

andrewnicols:
> Note: I'm completing the API checks for PHPUnit classes (that already have Namespace checks). Once working I plan to move them from being "phpunit only" to be "whole codebase" ones. Surely then we'll be adding more checks to the NamespaceStatement sniff added here.

Perfect - that was about what I had in mind too!

PR #63, opened 2023-09-18
Handle the case where an abstract testcase exists

andrewnicols:
Abstract methods were not handled properly. A method with no body has no scope_closer.

Discovered in a real world issue!


stronk7 (approved):
Good one! LGTM

PR #64, opened 2023-09-18
Abstract tests do not need coverage data

andrewnicols:
Coverage data for abstract classes is optional. In most cases it will make sense to belong to the concrete class that implements it.


andrewnicols:
Detected during a real review.

stronk7:
Hi,

I've taken the liberty of adding a new commit to your branch to ensure that the new code is covered (we already had a fixture to test for "skipped" classes).

Let's see how it ends... ciao :-)

stronk7 (approved):

PR #67, opened 2023-09-22
Test files must be named _test.php

andrewnicols:


stronk7:
Thanks @andrewnicols , I can confirm that the changes fix the problem with `coverage.php` being reported as missing any valid testcase class.

stronk7 (approved):

PR #69, opened 2023-09-24
Apply for export-ignore towards better release packages

stronk7:
When used as a dependency in other projects, we don't want the release packages to include any development file. This includes CI files, test files, docs...

This is useful to both:
- Reduce the package sizes noticeably.
- Ensure that composer, by default, (--prefer-dist) doesn't include them. If anybody prefers to get the complete version they always can use --prefer-source or clone the repository.

This is considered good practice, as far as there can be problems with tests and other stuff when the dependencies include everything.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,16 @@
+# Ignore all these files for release packages.
+# Using Composer `--prefer-dist` will not include these files.
+# Use `--prefer-source` to include them.
+# See https://blog.madewithlove.be/post/gitattributes/
+.gitattributes   export-ignore
+.gitignore       export-ignore
+.phplint.yml     export-ignore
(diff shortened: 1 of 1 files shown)


stronk7:
For the records, I've faced with this problem with this (draft) pull request:

https://github.com/moodlehq/moodle-local_codechecker/pull/221

There, you can see how `moodle-local_codechecker` tries to execute `moodle-cs` own tests when it's installed via Composer. Right now the only alternative that I've in the plugin is to add some `post-xxxx` Composer script proceeding to remove those files.

Ciao :-)

andrewnicols (approved):

stronk7:
Tatata!

PR #70, opened 2023-09-25
Exclude the PSR1.Classes.ClassDeclaration.MissingNamespace error

### LINKED ISSUE #65: moodle-extra requires class namespaces for mod_form and backup/restore
_Based on a comment here: https://github.com/moodlehq/moodle-cs/issues/26#issuecomment-1724302662_

The sniff `PSR1.Classes.ClassDeclaration.MissingNamespace` is reporting that "Each class must be in a namespace of at least one level (a top-level vendor name)", but Moodle still expects certain classes to exist in the global namespace, such as `mod_<modname>_mod_form` and the `{backup,restore}_<modname>_activity_task` classes. Can we exclude that rule for now until code supports namespaced versions of those classes? Or is it already possible to use namespaces classes and I just don't know?

stronk7:
Moodle doesn't mandate (or support) namespaces for all classes, so this PSR1, included by PSR12, included by moodle-extra, needs to be excluded. Soon we'll have proper/custom namespace sniffs.

Fixes #65

### DIFF
--- a/moodle-extra/ruleset.xml
+++ b/moodle-extra/ruleset.xml
@@ -23,6 +23,9 @@
         <!-- Moodle already defines its own line length, so remove this from the PSR-12 standard -->
         <exclude name="Generic.Files.LineLength.TooLong"/>
 
+        <!-- Moodle doesn't mandate (nor support) namespaces everywhere, so remove this for now -->
+        <exclude name="PSR1.Classes.ClassDeclaration.MissingNamespace" />
+
         <!-- Moodle has its own custom sniff for side effects  -->
         <exclude name="PSR1.Files.SideEffects.FoundWithSymbols"/>
 


moodle-extra/ruleset.xml
```diff
@@ -23,6 +23,9 @@
         <!-- Moodle already defines its own line length, so remove this from the PSR-12 standard -->
         <exclude name="Generic.Files.LineLength.TooLong"/>
 
+        <!-- Moodle doesn't mandate (neither support) namespaces everywhere, so remove this for now -->
```
jrchamp, line 26:
```suggestion
        <!-- Moodle doesn't mandate (nor support) namespaces everywhere, so remove this for now -->
```

jrchamp (commented):
Thank you! I'm so excited to start using moodle-extra 🥳

andrewnicols (approved):
I'd noticed this one too but hadn't gotten around to excluding it.

What I'd like to do is introduce a new sniff which reports missing or incorrect namespaces where the file is in a `classes` directory though.

stronk7:
> a new sniff which reports missing or incorrect namespaces where the file is in a classes

yes, yes, that's part of my namespaces sniff. just need the time to finish adding to them the new APIs stufff... but soon we'll have it. In any case... I don't think we will found many of them, it's sort of useless, because if it's incorrect... it won't autoload... but yes, good enough at least when coding and making typos.

Ciao :-)

PR #71, opened 2023-09-27
PHPUnit tests should have a return type

andrewnicols:
Generally speaking we shuld have a void return type on all tests in a testcase, but there may be other valid types if tests are   chained. We can't easily detect that, and other sniffs already detect if the return type is correct (not ours).

This sniff just checks for any test_ method which has no return type, and adds a `: void` if none is found. I don't think we need much more effort than that because 99% of our cases void is the correct value.

### DIFF
--- a/moodle/ruleset.xml
+++ b/moodle/ruleset.xml
@@ -110,6 +110,11 @@
     -->
     <rule ref="moodle.PHPUnit.TestCaseProvider"/>
 
+    <!--
+        Tests should have a return type. The default return type should be void.
+    -->
+    <rule ref="moodle.PHPUnit.TestReturnType"/>
+
     <!-- Disable this exact error unless it's approved -->
     <rule ref="moodle.Commenting.InlineComment.SpacingAfter">
         <severity>0</severity>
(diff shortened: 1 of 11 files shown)


andrewnicols:
Amended to be master (404) and up only :)

andrewnicols:
I agree we should make all unit test classes final.

moodle/ruleset.xml
```diff
@@ -110,6 +110,11 @@
     -->
     <rule ref="moodle.PHPUnit.TestCaseProvider"/>
 
+    <!--
+        Tests should have a return type. The default return type should be void.
+    -->
+    <rule ref="moodle.PHPUnit.TestReturnType"/>
```
stronk7, line 116: Fun one... I just realised... why do we need to add any "moodle.*" to the ruleset file? Only if there is some exception to their handling (say, disable, or lower from error to warning, or configure something...) they are needed.

It seems that we have missed that detail lately and there are 2-3 cases that can be safely removed.

Or am I missing anything?

stronk7:
I've created #74 about the `final` keyword, so we don't forget it.

stronk7:
And I've created #75 about to review `moodle/ruleset.xml` and remove `moodle.*` rules from it (unless they are changing something from defaults).

stronk7 (approved):
Given that the PR looks ok and the only point detected has beed reported as #75 , this can be safely approved. Thanks!

PR #73, opened 2023-10-13
Include checks for duplicate array keys

andrewnicols:
Discussed in MDL-79698.


paulholden (approved):
Thanks Andrew, just as we discussed - I think this is a good check to add

PR #80, opened 2023-11-14
Allow missing commas when last element is same line than closer

stronk7:
Comes with some tests to ensure that the case is now covered and we are allowing to miss those commas in the last array element when the array closer is in same line.

Fixes #76

PR #81, opened 2023-11-24
Allow missing commas when last element is same line than closer (take #2)

### LINKED ISSUE #76: NormalizedArrays.Arrays.CommaAfterLast sniff is wrong for Moodle
The newly added NormalizedArrays.Arrays.CommaAfterLast sniff gtom #60 does not correctly impleement Moodle coding style.

The good example

```php
$myarray = [1, 2, 3, 'Stuff', 'Here',
    $a, $b, $c, 56.44, $d, 500];
```

from https://moodledev.io/general/development/policies/codingstyle#numerically-indexed-arrays now gives a spurious failure.

stronk7:
This is a clone of #80 that was closed by mistake...

Comes with some tests to ensure that the case is now covered and we are allowing to miss those commas in the last array element when the array closer is in same line.

Fixes #76

Holding this as draft PR until [PHPCSExtra 1.2.0](https://github.com/PHPCSStandards/PHPCSExtra/milestone/13) is released. Then we'll have to change `composer.json` to point to that new release and this PR will be ready.

### DIFF
--- a/moodle/Sniffs/Arrays/CommaAfterLastSniff.php
+++ b/moodle/Sniffs/Arrays/CommaAfterLastSniff.php
@@ -1,0 +1,226 @@
+<?php
+/**
+ * PHPCSExtra, a collection of sniffs and standards for use with PHP_CodeSniffer.
+ *
+ * @package   PHPCSExtra
+ * @copyright 2020 PHPCSExtra Contributors
+ * @license   https://opensource.org/licenses/LGPL-3.0 LGPL3
+ * @link      https://github.com/PHPCSStandards/PHPCSExtra
+ */
+
+namespace MoodleHQ\MoodleCS\moodle\Sniffs\Arrays;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+use PHP_CodeSniffer\Util\Tokens;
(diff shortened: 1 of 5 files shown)


stronk7:
Note: I've added a candidate commit here that proceeds to:
- Copy the PHPCSExtra's CommaAfterLastSniff to our moodle standard.
- Amend tests, the ruleset.xml to use our copied Sniff instead of the upstream one.
- Make a note about that commit being reverted once PHPCSExtra releases v1.2.0 and we start using it (I will create issue if we agree that it's a good idea to advance with this alternative plan).

I've commented with the same @ #76.

(note that it's causing the code-coverage check to fail but we shouldn't mind too much. That is tested upstream in PHPCSExtra and here we are just testing the outcome for our very-own allowed multi-line arrays).

Ciao :-)

stronk7:
Ok, we are going ahead with this PR now. And #82 will be in charge of reverting commit 1c295bb0121247ca06ed4efd2d4e85d710da75d0 once the fixed upstream Sniff is released.

stronk7:
Self-merged. As commented above, the code-coverage report, missing 7 lines is ignorable, that Sniff is fully tested upstream and here we are only testing that the outcome matches Moodle coding style cases.

PR #83, opened 2023-12-02
Move phpcs from squizlabs to phpcsstandards

stronk7:
PHP_Codesniffer is being abandoned by squizlabs. See:

https://github.com/squizlabs/PHP_CodeSniffer/issues/3932 (old org) https://github.com/PHPCSStandards/PHP_CodeSniffer/issues/109 (new org)

So it has been moved (copied "manually", not transferred, not good, squizlabs) to the [PHPCSStandards](https://github.com/PHPCSStandards) organisation.

So this PR just moves us to start using the new organisation. Let's see how things evolve. It would be so great if the new one gets some support... (right now it's a 1-person project, basically).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
     ],
     "require": {
         "dealerdirect/phpcodesniffer-composer-installer": "^1.0.0",
-        "squizlabs/php_codesniffer": "^3.7.2",
+        "phpcsstandards/php_codesniffer": "^3.7.2",
         "phpcsstandards/phpcsextra": "^1.1.0",
         "phpcompatibility/php-compatibility": "dev-develop#0a17f9ed"
     },


jrchamp:
> So this PR just moves us to start using the new organisation. Let's see how things evolve. It would be so great if the new one gets some support... (right now it's a 1-person project, basically).

Any chance Martin could be convinced to sponsor? 😃  https://github.com/sponsors/PHPCSStandards  I'll definitely be asking my org if we can do something too.

junpataleta (approved):
Thanks, Eloy! Looks spot-on. Moving this forward!

jrfnl:
> I'm not sure to understand which is the advantage of keeping the old "organisation" name (non-matching the real one). 

Well, there are two advantages:
1. It saves 200.000+ projects the task of switching over.
    Just to give you some idea, we changed the organisation name of PHPCompatibility (from [`wimg`](https://packagist.org/packages/wimg/php-compatibility) to [`phpcompatibility`](https://packagist.org/packages/phpcompatibility/php-compatibility)) 5 years ago and there are _still_ packages which haven't switched over.
2. It will save me and maintainers of various external standards tons of support requests as the files in `vendor/bin` only get replaced when the `replace` directive is found (which would be in the 3.8.0 release) and don't necessarily get replaced when a custom `bin` directory is used, which could lead to worlds of pain.

stronk7:
Thanks @jrfnl , I'll continue asking @ https://github.com/PHPCSStandards/PHP_CodeSniffer/pull/135 , things are moving so quickly that better share things in the canonical issues and prs.

PR #86, opened 2023-12-19
GHA: Move from always() to !cancelled()

stronk7:
And add extra step to mark cancelled jobs as failed.

### DIFF
--- a/.github/workflows/phpcs.yml
+++ b/.github/workflows/phpcs.yml
@@ -33,24 +33,25 @@ jobs:
           composer install
 
       - name: Run phplint
-        if: ${{ always() }}
+        if: ${{ !cancelled() }}
         run: |
           ./vendor/bin/phplint
 
       - name: PHP Copy/Paste Detector
-        if: ${{ always() }}
+        if: ${{ !cancelled() }}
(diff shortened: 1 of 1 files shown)


stronk7:
Self-merging this, so further changes can happen...

stronk7:
Drat, it seems that the tests that were passing ok some weeks ago (have started to fail now with PHP 7.4, some updated component I bet.

Example: https://github.com/moodlehq/moodle-cs/actions/runs/7487886902/job/20381269027

Will look to this in separate issue and try to hot-fix it... ciao :-)

stronk7:
Just to confirm, it seems that the new https://github.com/nikic/PHP-Parser/releases/tag/v5.0.0 (released 4 days ago) is the change leading to the problems... so, apart from investigating and reporting it upstream... I'm going to try to force v4 here to be back to green ASAP.

stronk7:
And https://github.com/moodlehq/moodle-cs/pull/89 will fix the troubles, by making nikic/php-parser to stay sticky to v4 (curious that both phpunit and php-code-coverage have started to allow v5 when they are exactly the ones breaking with v5).

Issue #88, opened 2024-01-11
Allow nikic/php-parser v5 to be used once we move to php80 as requirement


stronk7:
It seems that both phpunit and php-code-coverage have
started to accept nikic/php-parser v5, but that's leading
to problems with PHP 7.4 runs, that we still support.

So, with this commit, we are locking nikic/php-parser v4: xxxxx

Once we officially make moodle-cs to only support php > 8.0
we can remove this lock (basically, remove the commit above).

This issue will be in charge of that.

PR #89, opened 2024-01-11
Lock nikic/php-parser v4

stronk7:
It seems that both phpunit and php-code-coverage have started to accept nikic/php-parser v5, but that's leading to problems with PHP 7.4 runs, that we still support.

See https://github.com/moodlehq/moodle-cs/actions/runs/7487886902 (that was passing 2-3 weeks ago).

So, with this commit, we are locking nikic/php-parser v4

Once we officially make moodle-cs to only support php >= 8.0 we can remove this lock. Issue: #88

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,6 +46,7 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^9.6",
+        "nikic/php-parser": "^4.18",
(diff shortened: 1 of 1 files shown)


andrewnicols (approved):

stronk7:
Tata!

PR #91, opened 2024-01-17
Check TODO and @todo comments looking for extra info

### LINKED ISSUE #90: Verify that all the "todo" comments have an MDL issue associated
Commented recently in a chat, it would be interesting to start emitting a warning when any "todo" comment:
- `// TODO: blah, blah, blah...`
- `@todo blah, blah, blah...`
is missing any MDL issue associated.

Surely we could apply for the same in @deprecated phpdoc tags too.

Note that all them are part of the coding style:

- https://moodledev.io/general/development/policies/codingstyle#using-todo
- https://moodledev.io/general/development/policies/codingstyle#deprecated-and-todo

Ciao :-)

stronk7:
This will look for some extra information to exist in:

- Inline TODO comments: // TODO blah, blah.
- PHPDoc @todo comments: /** @todo blah, blah. */

By default it will look for all TODOs having some 'MDL-[0-9]+' (regular expression), but it can be customised, in ruleset , phpcs.xml file or in CLI execution by setting the `$commentRequiredRegex` property or the `moodleTodoCommentRegex` configuration option  to any alternative regular expression.

Setting it to empty string will disable the sniff at all effects, useful for non-core stuff like plugins.

Fixes #90

### DIFF
--- a/moodle/Sniffs/Commenting/TodoCommentSniff.php
+++ b/moodle/Sniffs/Commenting/TodoCommentSniff.php
@@ -29,6 +29,7 @@ namespace MoodleHQ\MoodleCS\moodle\Sniffs\Commenting;
 
 // phpcs:disable moodle.NamingConventions
 
+use PHP_CodeSniffer\Config;
 use PHP_CodeSniffer\Files\File;
 use PHP_CodeSniffer\Sniffs\Sniff;
 
@@ -71,6 +72,11 @@ class TodoCommentSniff implements Sniff
      * @return void
      */
     public function process(File $phpcsFile, $stackPtr): void {
+        // If specified, get the regular expression from the config.
+        if (($regex = Config::getConfigData('moodleTodoCommentRegex')) !== null) {
+            $this->commentRequiredRegex = $regex;
+        }
+
         // If the regular expression is empty, then we don't want to do anything.
         if (empty($this->commentRequiredRegex)) {
             return;
--- a/moodle/Sniffs/Commenting/TodoCommentSniff.php
+++ b/moodle/Sniffs/Commenting/TodoCommentSniff.php
@@ -1,0 +1,180 @@
+<?php
+
+// This file is part of Moodle - https://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <https://www.gnu.org/licenses/>.
+
+/**
+ * Checks that some comments have a corresponding issue in the tracker.
+ *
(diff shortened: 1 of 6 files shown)


stronk7:
I've created these 2 issues (CiBoT and moodle-plugin-ci) to avoid this new sniff to be executed for plugins, it's only for core:

- CiBoT: https://github.com/moodlehq/moodle-local_ci/issues/289
- moodle-plugin-ci: https://github.com/moodlehq/moodle-plugin-ci/issues/266

moodle/Sniffs/Commenting/TodoCommentSniff.php
```diff
@@ -0,0 +1,180 @@
...
+     * then this Sniff will do nothing.
+     *
+     * Example values:
+     * - 'MDL-[0-9]+': The default value, a Moodle core issue is required.
+     * - 'CONTRIB-[0-9]+': A Moodle plugin issue is required.
+     * - '(MDL|CONTRIB)-[0-9]+': A Moodle core or plugin issue is required.
+     * - 'https://': Any URL is required.
+     * - '' (empty string or null): No check is done.
+     */
+    public ?string $commentRequiredRegex = 'MDL-[0-9]+';
```
kabalin, line 54: Since the package is used not only for MDL, thinking may be this needs to be empty, so that by default mode for this sniff is off? Then, projects maintaners who want to use it, have to define appropriate regex in `phpcs.xml`? Otherwise this becomes "action required" change in moodle-plugin-ci.

stronk7: Yeah, that's the reason I created the 2 issues commented above (in CiBoT and moodle-plugin-ci), to ensure that we disable it by default when those 2 tools run 'phpcs'.

I also considered to make "empty" the default, but was not convinced about that being the correct default (it's all about what's considered the normal use and what's considered the exceptional use, needing to adapt).

So I'm 65% about consider MDL-xxx the default (current status) and 35% about to switch to empty by default...

kabalin (approved):
OK, I agree to keep MDL default as nomal use case. Thanks for prompt solution for this Eloy!

kabalin (approved):

kabalin (approved):

paulholden (approved):

stronk7:
Ok, please don't merge this yet.

We need to change this Sniff `commentRequiredRegex` property to become a configuration option.

That way other tools can use `--runtime-set` (and others) in command line to get that aspect configured (without having to play with ruleset or phpcs.xml files).

Ciao :-)

stronk7:
Ok, I've added a new, 3rd commit, that adds the possibility of using a config setting to specify the desired regex to be used.

> Add a new `moodleTodoCommentRegex` config option
>
>    While everything continues working the same and the
>    Sniff property (`commentRequiredRegex`) still is available
>    and defaulting to `MDL-[0-9]+`, now we can, also, from command
>    line, decide which regular expression to use.
>  
>    That will make things easier from other tools, not having
>    to modify the ruleset phpcs.xml files. Instead just run:
>  
>    `vendor/bin/phpcs --runtime-set moodleTodoCommentRegex 'CONTRIB-[0-9]+'`
>  
>    or any other valid regular expression (https, github urls...)
>  
>    In order to disable the Sniff, you can pass to it an empty string
>    (apart from excluding it completely if desired, of course)
>  
>    `vendor/bin/phpcs --runtime-set moodleTodoCommentRegex ''`

paulholden (approved):
Thanks a lot @stronk7 super!

stronk7:
Thanks @paulholden and @kabalin !

PR #94, opened 2024-01-26
Avoid abstract infinite loop

### LINKED ISSUE #93: Codechecker hangs when parsing abstract functions
The codechecker hangs (apache process goes up to 100% CPU usage, will not stop until being killed by hand) when trying to parse a function in a test class of the following scheme:
`abstract function test_SOME_NAME_OF_THE_FUNCTION();`

It does not hang if the abstract function does not start with "test_" prefix.

stronk7:
Fix a couple of Sniffs that were able to enter into an infinite loop when processing `test_xxx` abstract methods.

Also adds a change to previous guard against that, because it was causing the abstract methods not to be processed.

Covered with tests.

Fixes #93


stronk7:
I'm self-merging this towards a round of releases soon.

PR #95, opened 2024-02-05
Enums do not have side effects

andrewnicols:
Add support for ENUMs to the MoodleInternal sniff


stronk7 (approved):
Looks perfect, and covered, hence, sold!

Thanks!

PR #100, opened 2024-02-13
Avoid PHP error with BoilerplateComment and empty files

### LINKED ISSUE #97: moodle.Files.BoilerplateComment fails on empty files
Where a file has only an opening `<?php` tag, the BoilerplateComment sniff fails with the error message: An error occurred during processing; checking has been aborted. The error message was: Undefined array key 1 in /path/to/vendor/moodlehq/moodle-cs/moodle/Sniffs/Files/BoilerplateCommentSniff.php on line 78. The error originated in the moodle.Files.BoilerplateComment sniff on line 78

I would expect this not to fail but to let the user know that the copyright statement is missing.

You could argue that files like this shouldn't exist at all, but there are cases where they're necessary, for example moodle-plugin-ci's checks on plugins fail if lib.php is missing but there is often no need for any content in this.

stronk7:
Covered with a specific new test.

Fixes #97


moodle/Tests/FilesBoilerPlateCommentTest.php
```diff
@@ -92,6 +92,19 @@ public function test_moodle_files_boilerplatecomment_short() {
         $this->verify_cs_results();
     }
 
+    public function test_moodle_files_boilerplatecomment_short_empty() {
+            $this->set_standard('moodle');
```
junpataleta, line 96: Thanks, @stronk7. This looks spot-on to me.  The indentation of this part seems to need some fixing, though.

junpataleta (changes requested):

moodle/Tests/FilesBoilerPlateCommentTest.php
stronk7: Whops, it should be fixed now. Thanks!

junpataleta (approved):
Thanks, @stronk7 ! Looking good now.

Issue #92, opened 2024-01-17
Consider creating a new "moodle-for-plugins" standard


stronk7:
Commented @ #91 maybe it would be nice to have a new standard, similar to the `moodle-extra` one for easier use by tools that are checking plugins and not core, namely CiBoT and moodle-plugin-ci.

That way, each time that we introduce a new Sniff in the "moodle" standard, if it's not suitable for plugins, we just can disable it in the new `moodle-for-plugins` standard and done. Or modify any behaviour (error vs warning), or whatever.

Right now, when those exceptions happen, we have to create issues like:
- CiBoT: https://github.com/moodlehq/moodle-local_ci/issues/289
- moodle-plugin-ci: https://github.com/moodlehq/moodle-plugin-ci/issues/266

And set a custom configuration there, that is slow and hard.

Instead, if we create the new standard for plugins, everything will work automagically, the tools just will use that standard and done.

Disclaimer: the `moodle-for-plugins` name is not a proposal, just the concept, heh.

jrchamp:
`moodle-plugins` makes a lot of sense to me. Right now, I'm using `moodle-extra` and like the idea of avoiding core-specific rules, so hopefully this plan will also create `moodle-plugins-extra` or `moodle-extra-plugins`.

andrewnicols:
Another option for this is to have an option set in the plugin's own `phpcs.xml`, for example:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<ruleset name="MoodleCore">
  <config name="moodle-type" value="plugin" /> 
</ruleset>
```

This can be accessed within the relevant sniffs:

```php
public function process(File $file, $pointer) {
    if ($file->config->getConfigData('moodle-type') === 'plugin') {
        return;
    }

    // ...
```

This has the benefit that the same standard sniffs can be used without doubling up.

stronk7:
Yeah, both are valid points,

- if there is a moodle-for-plugins, we'll need a moodle-for-plugins-extra.
- we can avoid the duplication by using some config setting controlling sniffs behaviour.

I think that having a (two) new standards is not much complex as far as they would include their parents and just enable/disable exceptional stuff.

In the other side, making it via configuration can be, also, a very good solution and the tools using phpcs (CiBoT, moodle-plugin-ci, codechecker, ... ) just have to apply for it or no.

[...]

PR #103, opened 2024-02-14
Unit test classes should be declared as either final or abstract

### LINKED ISSUE #102: New sniff: Non abstract test classes should be declared final
MDL-80927 is one example of what we want to fix.

We should make all test _classes_ (not cases) `final`, unless they are `abstract`.

andrewnicols:
Fixes #102 

The rationale for this change is based on the principle of least astonishment.

- If I modify a unit test, I expect it to only impact the thing that I expect it to test. If I add a new test, it should only test the thing I expect it to test.
- If someone is extending my unit test, then that is no longer the case, and I may be testing things which I did not intend, or over-testing the same thing.

To solve this the idea is:
- If I intend a test to be consumed, then I should make it abstract and I should extend it myself for my specific use-case.
- If I do not intend a test to be extended, then it should be declared final

### DIFF
--- a/moodle/Sniffs/PHPUnit/TestClassesFinalSniff.php
+++ b/moodle/Sniffs/PHPUnit/TestClassesFinalSniff.php
@@ -1,0 +1,124 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace MoodleHQ\MoodleCS\moodle\Sniffs\PHPUnit;
+
+// phpcs:disable moodle.NamingConventions
+
+use MoodleHQ\MoodleCS\moodle\Util\MoodleUtil;
+use PHP_CodeSniffer\Sniffs\Sniff;
+use PHP_CodeSniffer\Files\File;
+use PHPCSUtils\Utils\ObjectDeclarations;
+
+/**
+ * Checks that test classes are declared as final.
+ *
+ * @copyright  2024 Andrew Lyons <adrew@nicols.co.uk>
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class TestClassesFinalSniff implements Sniff {
+    public function register() {
+        return [
+            T_OPEN_TAG,
+        ];
+    }
+
+    public function process(File $file, $pointer) {
+        // If the file is not a unit test file, nothing to check.
+        if (!MoodleUtil::isUnitTest($file) && !MoodleUtil::isUnitTestRunning()) {
+            return; // @codeCoverageIgnore
+        }
+
+        // If we aren't checking Moodle 4.4dev (404) and up, nothing to check.
+        // Make and exception for codechecker phpunit tests, so they are run always.
+        if (!MoodleUtil::meetsMinimumMoodleVersion($file, 404) && !MoodleUtil::isUnitTestRunning()) {
+            return; // @codeCoverageIgnore
+        }
+
+        // Get the file tokens, for ease of use.
+        $tokens = $file->getTokens();
+
+        // We only want to do this once per file.
+        $prevopentag = $file->findPrevious(T_OPEN_TAG, $pointer - 1);
+        if ($prevopentag !== false) {
+            return; // @codeCoverageIgnore
+        }
+
+        $cStart = $pointer;
+        while ($cStart = $file->findNext(T_CLASS, $cStart + 1)) {
+            if (MoodleUtil::isUnitTestCaseClass($file, $cStart) === false) {
+                // This class does not relate to a unit test.
+                continue;
+            }
+            $className = ObjectDeclarations::getName($file, $cStart);
+            if (substr($className, -5) !== '_test') {
+                continue;
+            }
+
+            $classInfo = ObjectDeclarations::getClassProperties($file, $cStart);
+
+            if ($classInfo['is_final']) {
+                // Already final.
+                continue;
+            }
+
+            if ($classInfo['is_abstract']) {
+                // See if this class has any abstract methods.
+                $mStart = $cStart + 1;
+                $hasAbstractMethod = false;
+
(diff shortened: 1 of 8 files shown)


moodle/Sniffs/PHPUnit/TestClassesFinalSniff.php
```diff
@@ -0,0 +1,85 @@
...
+
+        $cStart = $pointer;
+        while ($cStart = $file->findNext(T_CLASS, $cStart + 1)) {
+            if (MoodleUtil::isUnitTestCaseClass($file, $cStart) === false) {
+                // This class does not relate to a unit test.
+                continue;
+            }
+            $className = ObjectDeclarations::getName($file, $cStart);
+
+            if (substr($className, -5) !== '_test') {
```
stronk7, line 63: Do we want this check in this case? That will ignore incorrect _testcase ones, isn’t it?

stronk7: I mean… all _testcase should be abstract, always. So they are subject to the logic/checks introduced here.

stronk7:
Apart from the detail in the review, surely the major "problem" of this nice Sniff is to determine how do we apply it (only for X.Y and up, warning and error in 1-year, only for non-plugins, ...).

As far as it's not part of the coding-style neither enforced by PHPUnit incoming versions... we need to be careful about it (back to the discussion happening @ #92).

Ciao :-)

andrewnicols:
Okay,

I've:
- split this into two sniffs:
  - one for tests which makes them final
  - one for testcases which makes them abstract
- applied it to 4.4 and up only

I think we need to add this to the epic for PHPUnit 10 as a part of that puzzle and add it to our coding style. This isn't something radical.

stronk7:
For fixtures...

`moodle/Tests/Sniffs/PHPUnit/fixtures`

or 

`moodle/Tests/fixtures/PHPUnit`

I've been doing the later lately and it seems that you've been doing the former. Of course, there are still a lot of fixtures to move so, at some time, we should decide which of the 2 fixtures dirs to use and go for it everywhere (for testing own Sniffs, other standards sniffs are another story).

moodle/Sniffs/PHPUnit/TestClassesFinalSniff.php
```diff
@@ -0,0 +1,126 @@
...
+                // See if this class has any abstract methods.
+                $mStart = $cStart + 1;
+                $hasAbstractMethod = false;
+
+                while ($mStart = $file->findNext(T_ABSTRACT, $mStart, $tokens[$cStart]['scope_closer']) !== false) {
+                    $hasAbstractMethod = true;
+                    break;
+                }
+                if ($hasAbstractMethod) {
+                    $file->addWarning(
```
stronk7, line 89: My brain is going to explode! Sure that there is a reason for the split of this and the `else` few lines below, but with the message being exactly the same, I'm missing it. Surely I need a siesta!

stronk7 (changes requested):
The final question in the review here is... do we want this to start warning now, with 4.4dev? Or want it for the next one?

moodle/Sniffs/PHPUnit/TestClassesFinalSniff.php
andrewnicols: If the test class is abstract, and contains abstract methods, then we cannot fix it automatically.
If the test class is abstract, but does not contains abstract methods, then we can make it final automatically (though it may well break something).

Really if they're abstract they should be converted to a testcase, but we can't automate that.

moodle/Tests/Sniffs/PHPUnit/TestcaseAbstractSniffTest.php
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Test the TestCasesAbstractSniff sniff.
+ *
+ * @package    local_codechecker
+ * @category   test
+ * @copyright  2024 Andrew Lyons <andrew@nicols.co.uk>
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ *
+ * @covers \MoodleHQ\MoodleCS\moodle\Sniffs\PHPUnit\TestClassesFinalSniff
```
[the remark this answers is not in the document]
andrewnicols: Yup - I suspected that from the failing coverage report. Will fix.

moodle/Tests/Sniffs/PHPUnit/TestClassesFinalSniffTest.php
```diff
@@ -0,0 +1,69 @@
...
+namespace MoodleHQ\MoodleCS\moodle\Tests\Sniffs\PHPUnit;
+
+use MoodleHQ\MoodleCS\moodle\Tests\MoodleCSBaseTestCase;
+
+// phpcs:disable moodle.NamingConventions
+
+/**
+ * Test the TestClassesFinalSniff sniff.
+ *
+ * @package    local_codechecker
```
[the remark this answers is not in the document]
andrewnicols: Done. Sold!!

moodle/Sniffs/PHPUnit/TestClassesFinalSniff.php
andrewnicols: Happy to make a change either way. It's pretty commonplace to have the same message for those that can be auto-fixed and those that cannot though from what I've seen.

andrewnicols:
> For fixtures...
> 
> `moodle/Tests/Sniffs/PHPUnit/fixtures`
> 
> or
> 
> `moodle/Tests/fixtures/PHPUnit`
> 
> I've been doing the later lately and it seems that you've been doing the former. Of course, there are still a lot of fixtures to move so, at some time, we should decide which of the 2 fixtures dirs to use and go for it everywhere (for testing own Sniffs, other standards sniffs are another story).

I've been doing the former because it keeps the fixtures closer to the thing under test, and makes them easier to find, but I'm happy to move them. Let me know your preference.

stronk7:
> I've been doing the former because it keeps the fixtures closer to the thing under test, and makes them easier to find, but I'm happy to move them. Let me know your preference.

Yeah, I think I agree with you, let's keep the fixtures closer to the the Test using them, specially now that we are creating subdirs for the tests.

Not in a hurry to move all current ones that have followed other paths, but let's go with your alternative for new ones.

moodle/Sniffs/PHPUnit/TestClassesFinalSniff.php
stronk7: Ok, I buy your argument, though there is still something that "resonates" in my brain (but sure that it's my brain's defect, so all ok!).

stronk7:
So, just to be 100% sure, as commented above:

- The final question in the review here is... do we want this to start warning now, with 4.4dev? Or want it for the next one?
- It will apply to all code checked (core and plugins).

I'm happy with that (4.4dev), will generate some noise for plugins, but it's a good thing, I think. And prepares us for the future, together with #106 and other details that have arrived recently (static providers, ....).

So, with your 👍 , agreement!

stronk7 (approved):

stronk7:
Sold!

PR #109, opened 2024-02-18
Bump squizlabs/php_codesniffer to 3.9.0

stronk7:
Also, bump PHPCompatibility to current dev version (e5cd2e24)

We were using version 9.3.5 that is from December 2019. Since then lots of new sniffs and compatibility with PHP 8.x versions have been provided.

So, while not ideal, let's use the dev version in the mean time.


andrewnicols (approved):

stronk7:
Merged, thanks!

PR #111, opened 2024-02-19
Lang string sorted please

### LINKED ISSUE #8: [CONTRIB-4147] Codechecker should check that strings in lang files are in alphbetical order
There may also be other common problems that we could auto-detect.
Note that local/codechecker/moodle/Sniffs/Files/LineLengthSniff.php is an example of a rule that does something different for lang files.

Reporter: Tim Hunt
Original issue: https://tracker.moodle.org/browse/CONTRIB-4147

stronk7:
Implement a new `moodle.Files.LangFilesOrdering` sniff in charge of check and fix string keys out of place.

- It recognises `// Deprecated since Moodle X.Y.X` comments and provides ordering within each of those "sections".
- If unknown comments are found, it stops providing fixes (for safety).
- Detects wrongly named variables (only `$string` is allowed).
- Detects wrongly constructed likes (having spaces within the brackets.
- Of course, detects incorrectly sorted string keys (warning) and, when possible, fix them.

Also, this moves the `MoodleUtil` tests to the new "fixtures" organisation that has been recently agreed (each test having its fixtures in same dir instead of global fixtures dir).

**TODO:**
- [ ] Cover the new Sniff with tests (I've them here semi-backed).
- [ ] Improve the fixer because, right now, it's using a simple swapping technique to end with the file sorted and, while working, it's not really performant (specially for big files with lots of unsorted stuff, like `lang/en/moodle.php` for example.

This fixes #8 

Created as draft, so we can start playing with it while fixing the points above.

Ciao :-)


marinaglancy:
Hi! Is this a check specific to the Moodle core only or available for the plugins as well? Because plugins may not have "Deprecated since Moodle..." they will have their own version, i.e. "Deprecated since 1.2.1" or something like this

stronk7:
Right now, any comment not matching "Deprecated since Moodle" does stop the fixer to work for any problem detected after it. But the checker works ok and will report keys out of order.

Not sure if changing the restriction to be just "Deprecated since" is a good idea... or it is...

stronk7:
Ok, I think this is now ready. The 2nd commit has all the important stuff. Take a look to the details there:

```
This sniff examines exclusively lang files, only for Moodle 4.4 and up,
and has the following features:
 - Detect "groups" of lang strings, (normal, deprecated) within the
   lang file, checking for correct order within each group.
 - If other comments, different from the deprecated ones, are found,
   for safety the sniff will stop fixing strings after that point,
   although it will continue reporting them as out of order.
 - Detect wrong variable names (different from `$string`).
 - Detect duplicate strings within each lang file group.
 - Detect strings with incorrect trailing contents after it (php code,
   invalid comment...).
 - Detect strings out of order within each lang file group and fix them.

Notes about the fixing:
 - Due to CodeSniffer fixing limits, this has been the really complex
   thing to implement and to guarantee that the fixer works also for
   "big" lang files (say error.php and similar ones).
 - To do so, instead of fixing every case 1 by 1, while we are
   checking the file... we have to accumulate all the changes
   in custom structure and then, apply everything together as
   a big changeset.
 - As said above, for safety, if an unexpected comment is found,
   the fixer stops applying any change after that line. In order to
   allow it to continue fixing, the comment must be removed. Only
   comments allowed are the "^// Deprecated since " ones, used to
   detect the groups of strings.
```

The trickiest part has been to implement the fixer, because, by default, CodeSniffer fixer limits were not allowing us to make the changes in a performant way, but in loops (controlled by `phpcbf` itself and limited to a max of 50 (not enough for our big files). That restriction has been finally avoided by accumulating all the changes in a unique big changeset and applying everything together.

I've tested it against some big lang files (moodle.php, error.php, ...) and it seems to be working ok.

moodle/Tests/Sniffs/Files/LangFilesOrderingTest.php
```diff
@@ -0,0 +1,97 @@
...
+                    33 => 'Unexpected string syntax, it should be',
+                    40 => 'The string key "yourself" is duplicated',
+                    42 => 'Unexpected string end',
+                ],
+            ],
+            'without strings' => [
+                            'lang/en/withoutLangStrings.php',
+                            [],
+                            [],
+                        ],
```
andrewnicols, line 89: Incorrect indenting.

andrewnicols (changes requested):
Had a first look through. Will go through in more detail tomorrow using xdebug to fully understand what's going on.

First glance looks good :)

stronk7:
Converted back to draft, while reviewing the results against `lang/en` I've detected a couple of border-cases that, while not the end of the world, better if we get them covered...

Thanks @andrewnicols!

moodle/Tests/Sniffs/Files/LangFilesOrderingTest.php
stronk7: Amended.

stronk7:
Ok, back to ready. Now there are only 2 files that cannot be fixed (`moodle.php` and `grades.php`) completely, and that's because, by coincidence, the very same line where they have an "illegal" (unexpected) comment, is also out of order.

When that happens (for example, it doesn't happen in `hub.php`, previously failing), the only solution is to present the 2 problems and let the user manually decide (or remove unexpected comment or move the string+comment to a correct place).

I'm creating a script to verify that we are not losing or modifying any string with the fixes, but I really think that we aren't, heh.

Ciao :-)

stronk7:
And, also for the records, checking the whole Moodle codebase, in < 5 minutes, got:

```
----------------------------------------------------------------------------------------------------------------------------------------
A TOTAL OF 2417 ERRORS WERE FIXED IN 341 FILES
----------------------------------------------------------------------------------------------------------------------------------------
PHPCBF FAILED TO FIX 2 FILES
----------------------------------------------------------------------------------------------------------------------------------------

Time: 4 mins, 25.94 secs; Memory: 278.01MB
```

(the 2 failed files are the ones commented above (`lang/en/grades.php`and `lang/en/moodle.php`, having both unexpected comment and wrong order in the same line).

andrewnicols (approved):

PR #112, opened 2024-02-19
Check category tag

andrewnicols:
Okay, so this adds a category check, which consumes the apis.json in the repo.

It currently doesnt' do anything fancy like checking the allowspread, or allowlevel2.

This is based on top of #110.


moodle/Sniffs/Commenting/CategorySniff.php
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * Processes php files and perform various checks with file.
+     *
+     * @param File $phpcsFile The file being scanned.
+     * @param int $stackPtr The position in the stack.
+     */
+    public function process(File $phpcsFile, $stackPtr) {
+        // The apis.json was introduced in Moodle 4.02.
+        // Make and exception for codechecker phpunit tests, so they are run always.
+        if (!MoodleUtil::meetsMinimumMoodleVersion($phpcsFile, 402) && !MoodleUtil::isUnitTestRunning()) {
```
stronk7, line 52: Do we really want this? The idea was to allow old Moodle branches to use the bundled `apis.json` file when not available in core, isn't it?

apis.json
```diff
@@ -0,0 +1,272 @@
+{
```
stronk7, line 1: Why are there 2 `apis.json` files?

moodle/Sniffs/Commenting/CategorySniff.php
andrewnicols: Damnit. I forgot to save the file before pushing it.

moodle/Tests/Sniffs/Commenting/CategorySniffTest.php
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * @covers \MoodleHQ\MoodleCS\moodle\Sniffs\Commenting\CategorySniff
+ */
+class CategorySniffTest extends MoodleCSBaseTestCase
+{
+
+    /**
+     * @dataProvider provider
+     */
+    public function test_from_provider(
```
stronk7, line 38: Maybe this is a too much generic name? better if it includes something like `test_category_correctness` (by similarity with the recent packages one) or so?

moodle/Util/MoodleUtil.php
```diff
@@ -16,6 +16,7 @@
 
 namespace MoodleHQ\MoodleCS\moodle\Util;
 
+use Exception;
```
stronk7, line 19: Not asking for any change on this, just sharing that it's the exact type of `use` use (heh), that always keep me thinking about what's better. To use or not to use (global scope classes), that's the question.

Again, not asking for changing it. At all!

stronk7 (approved):
I think this is 99% ready, just have added a couple of comments. Also you've already created #115 so, once the comments are resolved... feel free to merge this.

Ciao :-)

moodle/Util/MoodleUtil.php
andrewnicols: This is one of those cases of the IDE getting trigger happy. Fixed.

stronk7:
Yippee!

PR #113, opened 2024-02-23
Attributes should not be considered a side-effect

andrewnicols:
The use of Attributes should not impact the MOODLE_INTERNAL check.


stronk7:
Uh, oh. Maybe we have to instruct phplint to stop looking within any fixtures dir?

andrewnicols:
Huzzah - that did the trick. I hadn't realised we had the old fixtures excluded and I couldn't for the life of me work out why they were not failing.

stronk7 (approved):
Looks perfect, thanks!

PR #117, opened 2024-02-24
Define moodle-cs own standard, towards better coding

stronk7:
Here we are introducing the phpcs.xml.dist file that will guide our coding style within moodle-cs.

- It scans all the code but vendor and fixtures directories.
- It's the complete PSR12 standard but allowing single-line function declarations to have the opening curly bracket in the same line.
- With line length adjusted to 132 / 180.
- And a few more sniffs, not part of the PSR12, but good for readability (hopefully non-controversial):
  - && and || logical operators.
  - Short array syntax.
  - Space after commas.
  - Comma at the end of multi-line arrays.
  - PHPCompatibility standard to check for >=7.4 (with an exception about T_ENUM false positives).

And that's it!

Then, apart of the above, small adjustments to the composer.json and .gitignore files.

### DIFF
--- a/.github/workflows/phpcs.yml
+++ b/.github/workflows/phpcs.yml
@@ -41,6 +41,10 @@ jobs:
         if: ${{ !cancelled() }}
         run: ./vendor/bin/phpcpd --exclude moodle/Tests moodle
 
+      - name: Coding style
+        if: ${{ !cancelled() }}
+        run: ./vendor/bin/phpcs -ps .
+
       - name: Run phpunit
         if: ${{ !cancelled() }}
         run: |
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,21 @@
-vendor
+# Composer stuff
 composer.lock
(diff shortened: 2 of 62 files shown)


stronk7:
Created as draft, soon I'll add the BIG commit adapting all the code to the above.

Let's discuss if there is anything else to add to the configuration.

stronk7:
Ok, I've added 3 more commits, apart from the initial one, described above. This is a summary of all them:

1. Define the phpcs.xml.dist file, that configures the style to be used with `moodle-cs`.
2. Apply all the changes in code base towards being compliant with the style defined & agreed.
3. Modify the testing API to be, also compliant with the style.
4. Cover everything @ GHA with a new step that will check for compliance. Plus update changelog.

And that's all. I think this is now complete and can be considered for merging.

Ciao :-)

andrewnicols (approved):

stronk7:
Thanks!

PR #119, opened 2024-03-13
Add sniff to detect missing docblocks

andrewnicols:
Replaces:
```php
local_moodlecheck_registry::add_rule('filephpdocpresent')->set_callback('local_moodlecheck_filephpdocpresent');
local_moodlecheck_registry::add_rule('classesdocumented')->set_callback('local_moodlecheck_classesdocumented');
```

### DIFF
--- a/moodle/Sniffs/Commenting/MissingDocblockSniff.php
+++ b/moodle/Sniffs/Commenting/MissingDocblockSniff.php
@@ -1,0 +1,190 @@
+<?php
+
+// This file is part of Moodle - https://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANdTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <https://www.gnu.org/licenses/>.
+
+namespace MoodleHQ\MoodleCS\moodle\Sniffs\Commenting;
+
+use MoodleHQ\MoodleCS\moodle\Util\Docblocks;
+use MoodleHQ\MoodleCS\moodle\Util\MoodleUtil;
+use MoodleHQ\MoodleCS\moodle\Util\TokenUtil;
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+use PHP_CodeSniffer\Util\Tokens;
+use PHPCSUtils\Utils\ObjectDeclarations;
+
+/**
+ * Checks that all files an classes have appropriate docs.
+ *
+ * @copyright  2024 Andrew Lyons <andrew@nicols.co.uk>
+ * @license    https://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class MissingDocblockSniff implements Sniff
+{
+    /** @var array A list of standard method names used in unit test files */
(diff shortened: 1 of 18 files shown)


moodle/Util/Tokens.php
```diff
@@ -0,0 +1,61 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <https://www.gnu.org/licenses/>.
+
+namespace MoodleHQ\MoodleCS\moodle\Util;
+
+use PHP_CodeSniffer\Files\File;
+use PHPCSUtils\Utils\ObjectDeclarations;
+
+class Tokens
```
stronk7, line 23: Uhm, I think this is the only detail that I don't like too much. PHP_CodeSniffer already has `PHP_CodeSniffer\Util\Tokens` that is imported by various Sniffs as `Tokens`.
And that may cause confusion sometimes... maybe `TokenUtil` (coz it's similar to existing `MoodleUtil`) may be good enough to avoid the dupes?

moodle/Sniffs/Commenting/MissingDocblockSniff.php
```diff
@@ -0,0 +1,124 @@
...
+     * Processes php files and perform various checks with file.
+     *
+     * @param File $phpcsFile The file being scanned.
+     * @param int $stackPtr The position in the stack.
+     */
+    public function process(File $phpcsFile, $stackPtr) {
+        $tokens = $phpcsFile->getTokens();
+
+        // Each class, interface, trait, and enum must have a docblock.
+        // If a file has one class, interface, trait, or enum, the file docblock is optinoal.
```
stronk7, line 52: Typo: `optinoal`

moodle/Sniffs/Commenting/MissingDocblockSniff.php
```diff
@@ -0,0 +1,124 @@
...
+            $token = $tokens[$typePtr];
+            if ($token['code'] === T_FUNCTION && !empty($token['conditions'])) {
+                // Skip methods of classes, traits and interfaces.
+                continue;
+            }
+
+            $artefactCount++;
+
+            if ($docblock = Docblocks::getDocBlock($phpcsFile, $typePtr)) {
+                // There should be no empty lines between the artefact and the docblock.
```
stronk7, line 84: Side comment: For some reason we have been using `artifact` always. I see (didn't know) that `artefact` is the British work. Maybe we should move all cases to the later (not in this PR, just saying).

stronk7 (changes requested):
This looks great. Just to tiny comments and the `Tokens` name already in use thingy to solve.

stronk7:
One more little thing... this also replaces the "functionsdocumented" or no? I've seen that we skip it for methods, but it works for global scope functions correct?

moodle/Util/Tokens.php
andrewnicols: Good call. Done.

moodle/Sniffs/Commenting/MissingDocblockSniff.php
andrewnicols: Hah - taht explains a few things. I thought the i version were just typos. Happy to stay consistent for now and go to artifact.

moodle/Sniffs/Commenting/MissingDocblockSniff.php
andrewnicols: Fixed

stronk7 (approved):

stronk7:
Thanks @andrewnicols, it looks perfect, so I'm going to merge this now.

The only remaining point is to confirm if we should, also, remove (or no) the "`functionsdocumented`" check from `local_moodlecheck`, because in https://github.com/moodlehq/moodle-local_moodlecheck/pull/134 we only removed the file and class checks.

Ciao :-)

andrewnicols:
Sorry about that. I have raised https://github.com/moodlehq/moodle-local_moodlecheck/pull/137 for it.

PR #125, opened 2024-03-19
Add sniffs for file docblock @copyright and @license

andrewnicols:
Replaces https://github.com/moodlehq/moodle-local_moodlecheck/pull/139

### DIFF
--- a/moodle/Sniffs/Commenting/FileExpectedTagsSniff.php
+++ b/moodle/Sniffs/Commenting/FileExpectedTagsSniff.php
@@ -1,0 +1,160 @@
+<?php
+
+// This file is part of Moodle - https://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <https://www.gnu.org/licenses/>.
+
+namespace MoodleHQ\MoodleCS\moodle\Sniffs\Commenting;
+
+use MoodleHQ\MoodleCS\moodle\Util\Docblocks;
+use MoodleHQ\MoodleCS\moodle\Util\TokenUtil;
+use PHP_CodeSniffer\Config;
+use PHP_CodeSniffer\Sniffs\Sniff;
+use PHP_CodeSniffer\Files\File;
+use PHPCSUtils\Tokens\Collections;
+
+/**
+ * Checks that a file has appropriate tags in either the file, or single artefact block.
+ *
+ * @copyright  2024 Andrew Lyons <andrew@nicols.co.uk>
+ * @license    https://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class FileExpectedTagsSniff implements Sniff
+{
+    /**
+     * The regular expression used to match the expected license.
+     *
+     * Note that the regular expression is applied using preg_quote to escape as required.
+     *
+     * Note that, if the regular expression is the empty string,
+     * then this Sniff will do nothing.
+     *
+     * Example values:
+     * - Empty string or null: No check is done.
+     *  ''
+     * - The GNU GPL v3 or later license with either http or https license text
+     *   '@https?://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later@':
+     *
+     * @var null|string
+     */
+    public ?string $preferredLicenseRegex = '@https?://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later@';
+
+    /**
+     * Register for open tag (only process once per file).
+     */
+    public function register() {
+        return [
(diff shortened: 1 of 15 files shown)


moodle/Sniffs/Commenting/FileHasSniff.php
```diff
@@ -0,0 +1,148 @@
+<?php
```
stronk7, line 1: Isn't it name a little bit strange? Maybe 'FileExpectedTagsSniff' or 'FileHasTags' or ...

(not critical stuff in any case, for your consideration 100%).

moodle/Sniffs/Commenting/FileHasSniff.php
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * Checks that a file has appropriate tags in either the file, or single artefact block.
+ *
+ * @copyright  2024 Andrew Lyons <andrew@nicols.co.uk>
+ * @license    https://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class FileHasSniff implements Sniff
+{
+    /** @var string The license text expected */
+    protected static $licenseText = 'https://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later';
```
stronk7, line 37: Uhm... this can be problematic for plugins, that can use compatible licenses. And also within core, because there are lots of places pointing to "http" and, maybe also, old GPL2 or whatever.

Maybe this should default to empty, so no check is done, and later enable it if core execution is detected... 

Or, alternatively, I find it really similar to [what happened with the TODO/@todo sniff](https://github.com/moodlehq/moodle-cs/blob/main/moodle/Sniffs/Commenting/TodoCommentSniff.php#L36-L75), that we had to make it a configurable (public!) option, plus a config setting to allow 3rd executions to decide what to do. Maybe we can follow that very same approach.

stronk7 (changes requested):
Good one, just a couple of comments, with the one about how to handle the license text being the only one really important (or it will smash us later both with CiBoT and moodle-plugin-ci.

moodle/Sniffs/Commenting/FileHasSniff.php
andrewnicols: I like `FileExpectedTagsSniff`. Stolen :)

moodle/Sniffs/Commenting/FileHasSniff.php
andrewnicols: I had wondered the same myself but didn't quite come back to it.

I've gone with the same todo-like solution, but defaulting to:
```php
'@https?://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later@'
```

I'm also _not_ auto-fixing any license for now. Just too noisy.
```

File: moodle/Sniffs/Commenting/FileExpectedTagsSniff.php [hunk 3]
```diff
@@ -0,0 +1,160 @@
...
+     *
+     * Example values:
+     * - Empty string or null: No check is done.
+     *  ''
+     * - The GNU GPL v3 or later license with either http or https license text
+     *   '@https?://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later@':
+     *
+     * @var string
+     */
+    protected $preferredLicenseRegex = '@https?://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later@';
```
stronk7, line 51: Uhm, I think that this needs to be public if anybody wants to add a different license regex in their standard configuration. I may be wrong, but it sounds to me that I checked for examples when doing the TODO Sniff.

moodle/Sniffs/Commenting/FileExpectedTagsSniff.php
```diff
@@ -0,0 +1,160 @@
...
+                    'Missing @license tag',
+                    $stackPtr,
+                    'LicenseTagMissing'
+                );
+            }
+            return;
+        }
+
+        // If specified, get the regular expression from the config.
+        if (($regex = Config::getConfigData('moodleLicenseRegex')) !== null) {
```
stronk7, line 140: OT, and not needed for this PR, the other day I was reading that there is a way to document all these config options and also the configurable properties (in some XML format), that way when you look to the standard, all those options supported are shown.

I will create an issue about that once I find the details...

No action is needed here. Just sharing.

stronk7 (changes requested):
1 super tiny detail about the configurable property only!

moodle/Sniffs/Commenting/FileExpectedTagsSniff.php
stronk7: About the documentation... the current Sniff documentation system does not support documenting sniff properties or configuration options, just found it (only describing the Sniff behaviour is allowed). So, f we want to document all the options... surely we'll need to invent something (structured) or go the manual way (to /docs).

Reference: https://github.com/PHPCSStandards/PHP_CodeSniffer/issues/148#issuecomment-2007469426

Ciao :)

moodle/Sniffs/Commenting/FileExpectedTagsSniff.php
andrewnicols: Ah thanks. Fixed.

moodle/Sniffs/Commenting/FileExpectedTagsSniff.php
andrewnicols: I was thinking and wondering the same. We haven't even got anything in the readme yet.

andrewnicols:
Updated with latest changes (again) :)

stronk7 (approved):
Yay!

PR #127, opened 2024-03-20
Move unit test method Missing Docblock to new code

andrewnicols:
Either:
* Yes, it should
* Yes, but let's use a different Sniff response so it can be filtered and/or ignored in phpcs.xml
* No, it should not

All are very easy to implement.

I've gone for the middle ground for now.

### DIFF
--- a/moodle/Sniffs/Commenting/MissingDocblockSniff.php
+++ b/moodle/Sniffs/Commenting/MissingDocblockSniff.php
@@ -181,12 +181,14 @@ class MissingDocblockSniff implements Sniff
             $objectType = TokenUtil::getObjectType($phpcsFile, $typePtr);
 
             if ($isUnitTestFile) {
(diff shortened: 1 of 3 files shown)


stronk7:
Ho,

I'm not sure about this one... I don't think that we have required/suggested to document all the unit tests (functions) in the past (although I've not been able to find any reference in one - or the opposite - direction).

Uhm... impact out there can be certainly huge, if they were being skipped previously...

Not sure, not sure...

andrewnicols:
Bah - I think I missed a line in the old version. The old coding style made my brain completely miss that line, and there really weren't any good tests for the old one.

So we didn't previously add the error for test_ methods.

Added a commit to this patchset to do match that behaviour.

stronk7 (approved):
That looks perfect. I'll squash the 2 commits together, as far as they are highly coupled.

PR #128, opened 2024-03-20
Line length ignore for docs

andrewnicols:
Scratching an itch.


stronk7:
Heh,

from the examples in the PR, I feel that only the `@Given` has a pass.  The other 2 long cases (leaving apart the lang test that is unrelated) in phpdoc block and inline comment... I'd say that they should continue throwing warning/error. We don't want those long comments for any reason. Do we?

Maybe we can restrict the skipping to be a little more selecive (by tag, or by regex, ...).

Ciao :-)

andrewnicols:
We can... I'll update the patch some time later. Let's not block on this one. It's just one that does my head in!

PR #132, opened 2024-03-23
Fix docblocks before Multiline attributes

andrewnicols:
The line no was not updated to the line of the opening attribute.

Thanks to @ziegenberg for the test here.
Note: I had to amend it because PHP 7.4 cannot support multiline attributes, only single line (which are treated as comments)

Also please note that it is not necessary to sign-off your own commits. Typically the `Signed-off by` is used for peer code review.

### DIFF
--- a/moodle/Tests/Sniffs/Commenting/fixtures/MissingDocblock/docblock_with_multiline_attributes.php
+++ b/moodle/Tests/Sniffs/Commenting/fixtures/MissingDocblock/docblock_with_multiline_attributes.php
@@ -1,0 +1,50 @@
+<?php
+
+/**
+ * This file contains multiple testcases for multi line attributes.
+ */
+
+namespace MoodleHQ\MoodleCS\moodle\Tests\Sniffs\PHPUnit;
+
+defined('MOODLE_INTERNAL') || die(); // Make this always the 1st line in all CS fixtures.
+
+/**
+ * Example class. 
+ */
+#[\Attribute(
+    attr1: 'asdf',
+    attr2: 'asdf',
+)]
+class class_multiline_attribute {
+
+    /**
+     * Method attribute.
+     */
+    #[\Attribute(
+        attr1: 'asdf',
+        attr2: 'asdf',
+    )]
+    function method_multiline_attribute(): void {
+    }
+}
+
+/**
+ * Interface with multiline attributes.
+ */
+#[\Attribute(
+    attr1: 'asdf',
+    attr2: 'asdf',
+)]
+interface interface_multiline_attribute {
+}
+
+/**
+ * Trait with multiline attributes.
+ */
+
+#[\Attribute(
+    attr1: 'asdf',
+    attr2: 'asdf',
+)]
+trait trait_multiline_attribute {
+}
--- a/moodle/Tests/Sniffs/Commenting/fixtures/MissingDocblock/docblock_with_multiline_attributes.php
+++ b/moodle/Tests/Sniffs/Commenting/fixtures/MissingDocblock/docblock_with_multiline_attributes.php
@@ -41,10 +41,53 @@ interface interface_multiline_attribute {
 /**
  * Trait with multiline attributes.
  */
-
 #[\Attribute(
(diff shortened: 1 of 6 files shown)


moodle/Tests/Sniffs/Commenting/fixtures/MissingDocblock/docblock_with_multiline_attributes.php
```diff
@@ -0,0 +1,93 @@
+<?php
+
+/**
+ * This file contains multiple testcases for multi line attributes.
+ */
+
+namespace MoodleHQ\MoodleCS\moodle\Tests\Sniffs\PHPUnit;
+
+defined('MOODLE_INTERNAL') || die(); // Make this always the 1st line in all CS fixtures.
```
stronk7, line 9: Maybe, we can create an issue to get rid of these. I think that we don't need them anymore in the context of `moodle-cs` as a standalone product.

andrewnicols:
I wholly agree but let’s do it as a separate PR. I saw this one and did go
and check and there’s a mix of both

moodle/Tests/Sniffs/Commenting/fixtures/MissingDocblock/docblock_with_multiline_attributes.php
```diff
@@ -0,0 +1,93 @@
...
+ * Trait with multiline attributes.
+ */
+#[\Attribute(
+    attr1: 'asdf',
+    attr2: 'asdf',
+)]
+trait trait_multiline_attribute {
+}
+
+/**
```
stronk7, line 51: Do we have any example having a blank line between the end of attribute and artifact declaration? I imagine that it will lead to "not-documented" and/or "attribute-without-target", but from this sniff perspective, maybe it also makes sense to have it covered.

Example:

```
/**
 * Trait with multiline attributes.
 */
#[\Attribute(
    attr1: 'asdf',
    attr2: 'asdf',
)]

trait trait_multiline_attribute {
}
```

stronk7 (approved):
This looks ok, just have left a couple of independent comments to be considered. Feel free to merge after considering them, specially the separated-attributes one.

andrewnicols:
It isn’t valid but I don’t think we have an example. Can put one together.
Won’t be home for a few hours so don’t know if you want to add one and test
in my absence?

moodle/Tests/Sniffs/Commenting/fixtures/MissingDocblock/docblock_with_multiline_attributes.php
stronk7: I've created #135 about to remove those old `MOODLE_INTERNAL` lines.

moodle/Tests/Sniffs/Commenting/fixtures/MissingDocblock/docblock_with_multiline_attributes.php
stronk7: Ok, was looking to existing ' moodle.Commenting.MissingDocblock` tests, and, it seems that, for classes, we have that case already covered by the `class_only_with_attributes_incorrect_whitespace` fixture.

What I'm going to do is:
1. Add similar cases for method, interface and trait.
2. In the `` fixture, change all the ocurrences of `#[\Attribute(` to `#[someattribute(` because the former is incorrect.

Will add commit and, if everything looks ok, proceed to merge this.

stronk7: Done, I've added a 4th commit with the points commented above.

stronk7:
Done:
- #135 has been created about the `MOODLE_INTERNAL` removals.
- A 4th commit has been added here, covering the whitespace after attributes in interfaces, traits and methods (plus a small cleanup).

stronk7:
My side, this is ready now. I'll leave you, @andrewnicols , to make the merging honours if everything looks ok for you.

Ciao :-)

PR #134, opened 2024-03-23
Add sniff to check for first line descriptions

andrewnicols:
Replaces:
* phpdocsfistline
* functiondescription

Replaces https://github.com/moodlehq/moodle-local_moodlecheck/pull/142

### DIFF
--- a/moodle/Sniffs/Commenting/DocblockDescriptionSniff.php
+++ b/moodle/Sniffs/Commenting/DocblockDescriptionSniff.php
@@ -1,0 +1,102 @@
+<?php
+
+// This file is part of Moodle - https://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANdTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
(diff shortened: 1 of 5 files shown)


stronk7:
Q: Does this cover cases like:

```
/**
 * This is the first line of the block.
 * But the second an following aren't separated enough and
 * make the block to look weird.
 */
<<files, classes, ..., but methods, that I think don't require it, go here >>
```

I mean, it sounds to me that the 1st line alone (with empty line after), has been always a requirement for them. But we haven't enforced it, I think. And this may be the perfect sniff for that.

andrewnicols:
> Q: Does this cover cases like:
> 
> ```
> /**
>  * This is the first line of the block.
>  * But the second an following aren't separated enough and
>  * make the block to look weird.
>  */
> <<files, classes, ..., but methods, that I think don't require it, go here >>
> ```
> 
> I mean, it sounds to me that the 1st line alone (with empty line after), has been always a requirement for them. But we haven't enforced it, I think. And this may be the perfect sniff for that.

It doesn't cover that case right now, but it's not hard to add. Still, I'd prefer to do it as a separate PR in this case because there are a couple of ways to do it.

stronk7:
> Still, I'd prefer to do it as a separate PR in this case because there are a couple of ways to do it.

Agreed, we can create a new issue with some thoughts about what we may need to check:
- Blank line between summary (short desc) and description (and for which elements).
- Blank like between descriptions and tags.
- Maybe some ordering and spacing between tag "groups" (copyright, author, license => params => return => other tags)
- ....
(I'd look to what the phpdoc proposal says for anything not explicitly defined in our guidelines)

I'll proceed to review this soon... ciao :-)

stronk7:
Have created #136 to have the discussion above registered.

stronk7 (approved):

PR #140, opened 2024-03-28
Deprecated items do not need a description

andrewnicols:
Raised by @ziegenberg 

Note: I've included an extra commit her eto slightly refactor the Docblocks util to use the docblockPtr rather than find it from an unrelated pointer. This is much more efficient and I should have done this originally.

### DIFF
--- a/moodle/Sniffs/Commenting/CategorySniff.php
+++ b/moodle/Sniffs/Commenting/CategorySniff.php
@@ -46,15 +46,21 @@ class CategorySniff implements Sniff
      * @param int $stackPtr The position in the stack.
      */
     public function process(File $phpcsFile, $stackPtr) {
-        $categoryTokens = Docblocks::getMatchingDocTags($phpcsFile, $stackPtr, '@category');
+        $docPtr = Docblocks::getDocBlockPointer($phpcsFile, $stackPtr);
+        if (empty($docPtr)) {
+            // It should not be possible to reach this line. It is a safety check.
+            return; // @codeCoverageIgnore
+        }
+
+        $categoryTokens = Docblocks::getMatchingDocTags($phpcsFile, $docPtr, '@category');
         if (empty($categoryTokens)) {
             return;
         }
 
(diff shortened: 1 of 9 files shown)


moodle/Sniffs/Commenting/CategorySniff.php
```diff
@@ -46,15 +46,20 @@ public function register() {
      * @param int $stackPtr The position in the stack.
      */
     public function process(File $phpcsFile, $stackPtr) {
-        $categoryTokens = Docblocks::getMatchingDocTags($phpcsFile, $stackPtr, '@category');
+        $docPtr = Docblocks::getDocBlockPointer($phpcsFile, $stackPtr);
```
stronk7, line 49: Uhm... haven't checked but... what should this line return if there is a file with ZERO PHPDcoc blocks?

I was adding a new extra fixture (and moving them to "Category" dir) that is a file with just one class and without any PHPDoc block (aiming to cover that "return" that is not covered) and, surprise, surprise... when I run the test, the line continues being not-covered, aka, it's returning something no matter there isn't any block.

stronk7: Or, maybe, with other words, as far as the Sniff in only looking for `T_DOC_COMMENT_OPEN_TAG`, then that line is 100% superfluous/unreachable in that case. It's impossible for the function to return null when the pointer is a `T_DOC_COMMENT_OPEN_TAG`.

So, TLDR; or we delete it, or, if we keep it as guardian for any future change in the registered tokens, we `@codeCoverageIgnore` it. I'm happy either way.

andrewnicols: It's intended as a safety check, so yes we could/should add the codeCoverageIgnore.

This change has no actual change - I've just broken the pointer fetch to make it more explicit.
I'll ignore the coverage there.

stronk7 (approved):

PR #144, opened 2024-03-31
Properties should not be tested for regular constructors (Fixes #142)

### LINKED ISSUE #142: Problem with moodle.Commenting.VariableComment and constructor methods
It seems that `__construct()` methods (and maybe others) are reporting as missing all the params, no matter they are properly documented.

construct_problems

Just renaming the method to anything else (without `__`) makes the sniff to pass.

Ciao :-)

andrewnicols:
Only properties which define a visibility and therefore are promoted constructor properties are required to define a variable docblock.


stronk7 (approved):
Thanks for the quick fix, @andrewnicols . Looks perfect!

PR #147, opened 2024-04-02
Object names may not be present for all objects

andrewnicols:
Reported by @ewallah via commit comment here: https://github.com/moodlehq/moodle-cs/commit/237cf60e5157e7cee3b624ae1bd25df13fb0da14#commitcomment-140496069


sarjona (approved):
Hi @andrewnicols!
Thanks for fixing this issue. The patch makes sense, so I'm happy to approve and merge it :-)

PR #156, opened 2024-04-17
Make MissingDoclock code more specific (Fixes #154)

### LINKED ISSUE #154: Make MissingDocblock names more specific
At the moment the warnings/errors all use the identifier `Missing`. These should include the type that is missing, for example:

- `moodle.Commenting.MissingDocblock.File`
- `moodle.Commenting.MissingDocblock.Class`
- `moodle.Commenting.MissingDocblock.Method`
- `moodle.Commenting.MissingDocblock.Function`
- `moodle.Commenting.MissingDocblock.Constant`
- `moodle.Commenting.MissingDocblock.Property`

andrewnicols:


stronk7 (approved):
This looks good to me... I just have one suggestion...

Maybe it wouldn't be crazy to amend some of the related tests expectations and ensure that the new "codes" are being returned properly?

andrewnicols:
Agree - I'll try and set some time aside to do this.

stronk7:
Hi,

I've resolved conflicts here, added some specific code tests and amended the sniff slightly to ensure that the 1st char in the code is always upper (that is good practice, I think).

Will be merging this soon...

Issue #27, opened 2021-03-22
Adhere to PSR-12 rules for `declare`


andrewnicols:
https://www.php-fig.org/psr/psr-12/#3-declare-statements-namespace-and-import-statements

Specifically:
1. no spaces around the =
2. must be first item in the file

> Declare statements MUST contain no spaces and MUST be exactly declare(strict_types=1) (with an optional semi-colon terminator).
> 
> Block declare statements are allowed and MUST be formatted as below. Note position of braces and spacing:
> 
> ```
>  declare(ticks=1) {
>     // some code
> }
> ```

andrewnicols:
Closing as fixed by #57 which included the PSR-12, which includes these rules.

PR #158, opened 2024-04-21
Update BoilerplateCommentSniff and add fixes.

micaherne:
There are a few troublesome things I find about BoilerplateCommentSniff:

1. It doesn't allow you to use `declare(strict_types=1)` in the standard place - on the same line as the opening PHP tag. You have to put it directly after the Moodle comment, which is obviously valid but it's non-standard and harder to spot.
2. If you have anything at all on the line after the PHP tag, except for the Moodle boilerplate, you get 14 separate messages saying that some random token doesn't match a line of the Moodle boilerplate. I'd expect a single message saying that the boilerplate wasn't found, or at least for it only to attempt to validate comment tokens.
3. There are no fixes available

(I appreciate that the second point probably doesn't affect core development so much because all core files have the boilerplate, but it just swamps the output with noise if you have a plugin that doesn't yet have the boilerplate in any of the files.)

Could I propose this patch to improve things a bit? It:

* allows `declare()` calls on the same line as the opening PHP tag
* only checks the boilerplate for contiguous comment tokens, and bails out if it encounters any other kind of token
* provides fixes to automate inserting / fixing the boilerplate when it's found in the wrong place or incomplete

### DIFF
--- a/moodle/Sniffs/Files/BoilerplateCommentSniff.php
+++ b/moodle/Sniffs/Files/BoilerplateCommentSniff.php
@@ -30,7 +30,7 @@ use PHP_CodeSniffer\Util\Tokens;
 
 class BoilerplateCommentSniff implements Sniff
 {
-    protected static $comment = [
+    protected static array $comment = [
         "// This file is part of",
         "//",
         "// Moodle is free software: you can redistribute it and/or modify",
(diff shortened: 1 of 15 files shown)


moodle/Sniffs/Files/BoilerplateCommentSniff.php
```diff
@@ -64,53 +64,131 @@ public function process(File $file, $stackptr) {
...
+                            $stackptr,
+                            'InvalidTrailingCode'
+                        );
+
+                        return;
+                    }
+                }
+            }
+        }
+
```
andrewnicols, line 109: This stuff doesn't really belong in this sniff. It's not a part of the boilerplate, but a totally different sniff.
I think most of what you have here is covered by https://github.com/PHPCSStandards/PHP_CodeSniffer/blob/master/src/Standards/PSR12/Sniffs/Files/DeclareStatementSniff.php

moodle/Sniffs/Files/BoilerplateCommentSniff.php
```diff
@@ -64,53 +64,131 @@ public function process(File $file, $stackptr) {
...
         // Allow T_PHPCS_XXX comment annotations in the first line (skip them).
         if ($commentptr = $file->findNext(Tokens::$phpcsCommentTokens, $stackptr + 1, $stackptr + 3)) {
             $stackptr = $commentptr;
         }
 
-        // Find count the number of newlines after the opening <?PHP. We only
-        // count enough to see if the number is right.
-        // Note that the opening PHP tag includes one newline.
-        $numnewlines = 0;
-        for ($i = $stackptr + 1; $i <= $stackptr + 5; ++$i) {
```
andrewnicols, line 76: There's a few ways of doing this... see https://github.com/PHPCSStandards/PHP_CodeSniffer/blob/master/src/Standards/PEAR/Sniffs/Commenting/FileCommentSniff.php#L95-L104 for an example

andrewnicols (changes requested):

stronk7:
Hi,

just sharing... also note that, regarding `declare()` statements, we recently adhered about that to [PSR-12](https://www.php-fig.org/psr/psr-12/#3-declare-statements-namespace-and-import-statements) rules (in #27), so it's not clear for me why should we allow it in the php opening tag (if I've understood what is being proposed here).

Ciao :-)

micaherne:
Thanks for the feedback @andrewnicols and @stronk7! 

I see declare strict types in line with the PHP opening tag an awful lot, so I'd assumed it was standard practice, but if it's not allowed by the standard that's good. I've excised all the code relating to declares and it's simplified the patch a lot.

I've also added a unit test for the new error condition I added (CommentEndedTooSoon). I'd like to add tests for the fixes but I'm at a bit of a loss as to how to do that, as I can't find any examples anywhere - I've looked in quite a few standards.

micaherne:
Also I have a couple of questions that could help me to improve the patch if I knew the answers.

1. I retained the original behaviour where it will only check the boilerplate if it's found after five or fewer blank lines at the start of the file. I'm guessing this was for performance reasons, but it would make things a lot simpler and more logical if we could just find the first comment in the file with `$file->findNext(T_COMMENT, ...)` and check it for correctness regardless of where it is. Obviously for a big file with no comments this could mean searching through the whole file (although I'm pretty sure that would be really fast). Would that be acceptable or is only checking the first five tokens a hard requirement?

2. The text for checking the comment is missing the text "Moodle - https://moodle.org/" at the end of the first line. Reading between the lines this looks like it's so that the web URL can be replaced with a different one (given that any product name is supported, not just Moodle). Is it still a requirement to support other names and URLs or could this just be the canonical Moodle boilerplate text?

(I was also wondering whether the http: URLs could be changed to https: ?)

Do you know the answers to these?

micaherne:
I've got a version of this that I'm happy with now, I hope you might consider it. Apologies for the confusion around the declare statements - I was mistaken about that and I've removed all traces of it.

Hopefully it's clear enough from the code but the logic of the checks is now:

1. Check the first line of the first comment in the file (if it exists) matches "This file is part of". Otherwise assume there is no boilerplate, show an error, and stop checking.
2. Check each following "//" comment line matches the correct boilerplate text, and show an error if not. If there are too few contiguous comment lines, show an error saying that the boilerplate is incomplete and stop.
3. Check that the boilerplate is at the opening line of the file and show an error if it is not. (This will still show first even though it is checked last.)
4. Check that there is exactly one newline (or the end of the file) after the boilerplate.

(The last one is a new rule, but it seems reasonable to me and helps with the fixes. I'll remove it if it's not acceptable.)

The way the checks are done simplifies the corresponding fixes:

1. Insert the boilerplate at the first line.
2. Fix incorrect lines / complete with missing lines
3. Move boilerplate to first line of file
4. Remove multiple trailing lines

As well as the unit tests I ran the fixes against Moodle core and can't see any issues: https://github.com/moodle/moodle/compare/main...micaherne:moodle:boilerplate-sniff-test (apart from where there are some existing very non-standard copyright notices that couldn't be detected!)

stronk7:
Thanks for all the effort put on this, @micaherne.

I'm going to take a look to it now, sorry for the delay! Will come with some comments or suggestions if I find something.

Ciao :-)

stronk7:
FYI, I've taken the liberty of rebasing your commit on top of "main", to have it updated.

stronk7:
And another small edition, simply changing a few var names, no change in logic, so far.

stronk7:
And, now, I've added a good number of "fixture.php.fixed" files to ensure that the fixer is doing its job with all them.

Also, have added one new test (trailing_whitespace_missing) to complete the coverage.

stronk7 (approved):
Ok, I think this can be considered ready, with tests passing and all the previous discussions already resolved, hence, accepting.

micaherne:
@stronk7 thanks for fixing this up and getting it into production!

stronk7:
> @stronk7 thanks for fixing this up and getting it into production!

ty!

stronk7:
For the records, #168 has been created, regression of this, about we needing, surely, to allow "extra lines" in the boilerplate to exist, and only apply the `SingleTrailingNewLine` detection to the end of the whole thing.

PR #160, opened 2024-04-23
Do not warn about missing docblocks when a method has the #[\Override] attribute

### LINKED ISSUE #155: Train moodle.Commenting.MissingDocblock.Missing to understand #[\Override]
Moodle coding style rules say:

1. A method MUST have a PHP doc comment ...
2. Unless it is an overridden method, in which case it SHOULD NOT have a PHPdoc comment (because that is bad duplication).

Our code-checker rules have never understood this, and always given false-positives.

But, PHP 8.3+ now provides the #[\Override] annotation (See https://tracker.moodle.org/browse/MDL-79901 for discussion of the use of that in Moodle.) And that can safely be used in older PHP without errors.

So, we should teaching moodle.Commenting.MissingDocblock.Missing to use this annotation, and for such methods, give a warning if the comment is present (rather than if it is absent).

andrewnicols:
Fixes #155

POC

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format of this change log follows the advice given at [Keep a CHANGELOG](htt
 ## [Unreleased]
 ### Changed
 - Update composer dependencies to current versions, notably `PHP_CodeSniffer` (3.10.1) and `PHPCompatibility` (96072c30).
+- The `moodle.Commenting.MissingDocblock` sniff will now detect use of the Override attribute (Fixes #155).
 
 ## [v3.4.6] - 2024-04-03
 ### Fixed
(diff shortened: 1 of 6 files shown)


andrewnicols:
Need to add some unit tests for attributes helper.

I'm not going to make this handle fully-qualified checks right now. We'll see if phpcsextra or phpcs do so.

andrewnicols:
I've been running this locally for the past month and haven't encountered any weird bugs or issues. Pushing for review.

stronk7 (approved):
LGTM, I've left a comment @ #155 about we needing to add this to the devdocs and, maybe, explain better when inherited docs should be added or avoided.

https://moodledev.io/general/development/policies/codingstyle#functions

Ciao :-)

stronk7:
(whops, if you could rebase this on top of current main, that would be great, I'll do in a couple of hours if you cannot)

PR #161, opened 2024-05-13
Better integration tests, with correct exit status checks

stronk7:
There were a few problems with the integration checks, not properly detecting & asserting the expected exit statuses. Plus some problem using `runner.os` that uses different names (`Linux`, `Windows`, `macOS`), so switching to `matrix.os`

Now they are better checked, with some output added to explain the problem when the script runs.


andrewnicols (approved):

stronk7:
Tata!

PR #162, opened 2024-05-20
Bump to phpcs 3.10.0

stronk7:
And also to current dev phpcompatibility 96072c30

PR #165, opened 2024-05-28
New sniff to detect correct setUp/tearDown parent calls

### LINKED ISSUE #106: Detect that setUpBeforeClass, setUp, tearDown, and tearDownAfterClass are always calling to parent
While it's not critical right now, we'll need to ensure that all the chain of parents are always executed, else, we may be missing cleaning stuff.

More important, with PHPUnit 10 we MUST (are forced) to move big chunks of stuff from current `runBare()` method (that has been declared `final`, hence we cannot overwrite it anymore) to those methods. So better we go checking that all them are properly chained ASAP.

Note (copied from another project): for setup ones, the call normally goes at the beginning, and for tear down ones, it normally goes at the end, but having it anywhere is better than not having at all.

Ciao :-)

stronk7:
It detects various situations like:
  - Missing calls to parent.
  - Dupe calls.
  - Incorrect calls.

And, when possible, tries to fix the missing ones (not others) by adding them in a correct place:
  - For setUp cases, at the beginning of the method (after global and require statements, if any).
  - For tearDown cases, at the end of the method.

Only for 405 and above.

Of course, covered with tests.

Fixes #106

Note that this is built on top of #162.


stronk7:
These are the results when executing it against core (4.5dev). 492 cases detected and fixed.

https://github.com/stronk7/moodle/commit/c22f9e08f88c643bb3d56c694df9fb76ce233f96

Ciao :-)

moodle/Sniffs/PHPUnit/ParentSetUpTearDownSniff.php
```diff
@@ -0,0 +1,222 @@
...
+                        end($correctParentCalls),
+                        'Multiple' . ucfirst($method),
+                        [$method, $method]
+                    );
+                }
+
+                // If there are no calls to correct parent, report it. Fixable.
+                if (count($correctParentCalls) === 0) {
+                    // Any weird case where the method is empty, we need at very least 1 line. Skip.
+                    $startLine = $tokens[$tokens[$mStart]['scope_opener']]['line'];
```
andrewnicols, line 149: I'm not sure I agree that this is the right approach.

If the method is empty shouldn't we warn about it because it will block our future reset. I feel the correct fix is instead to remove these empty methods.

andrewnicols: The same applies to multi-line empty methods.

andrewnicols (changes requested):

moodle/Sniffs/PHPUnit/ParentSetUpTearDownSniff.php
```diff
@@ -0,0 +1,222 @@
...
+                    $endLine = $tokens[$tokens[$mStart]['scope_closer']]['line'];
+                    if ($startLine === $endLine) {
```
andrewnicols, line 151: Rather than this, perhaps something like:

```php
// Find the next thing that is not an empty token.
$ignore = \PHP_CodeSniffer\Util\Tokens::$emptyTokens;

$nextValidStatement = $phpcsFile->findNext($ignore, $mStart, $tokens[$mStart]['scope_closer'], true);
if ($nextValidStatement === false) {
    continue;
}
```

moodle/Sniffs/PHPUnit/ParentSetUpTearDownSniff.php
stronk7: Well, my assumption was that it makes no sense to have empty setup/teardown methods, no matter if there are sameline or multiline. Why would anybody ever do that?

I explicitly prevented those having same start and end line, that, again, is extremely unlikely case, aka:

```
public function setUp() { }
```

Because the fixer implementation does not work with them. As simple as that. So I decided that it was better to skip them completely than to report and fix them.

If you think that it's worth adding a new errorcode for them, I'm happy doing so. As said, I just imagined that they were not going to happen "ever" (near) and to skip them was enough.

About multiline-empty methods, really it's similar, who is going to write an empty multiline method? And, in case it does... what do we consider empty? Do we accept comment contents, for example? 

Again, it can be done (if the method only has whitespace and comments, suggest to remove it completely). Although current approach, that is to add the call to parent... is also correct IMO.

moodle/Sniffs/PHPUnit/ParentSetUpTearDownSniff.php
stronk7: Okidoki, I like it. Will apply soon.

stronk7:
Thanks @andrewnicols , I've cherry-picked your commit here, and applied a few small (comments, problem cases) modifications to it, to better understand what's going with those empty methods.

I think this now is perfect! 😎

andrewnicols (approved):

andrewnicols (approved):

PR #167, opened 2024-06-10
Detect multiple @coversDefaultClass uses in a class

### LINKED ISSUE #166: Make the TestCaseCovers sniff to detect multiple @coversDefaultClass cases
Recently I was playing with code-coverage generation and when running PHPUnit it was detected that we have a few unit tests having multiple `@coversDefaultClass` annotations.

That's plain wrong, only one is allowed (in fact PHPUnit stops when that happens). Such an annotation is about to define the default class to apply when the `@covers` annotations do no specify a class. Just that.

So we should start erroring in the Sniff when multiple are detected. Of course, only that annotation, classes and methods are allowed to have multiple `@covers` ones.

Ciao :-)

stronk7:
Pretty simple. Implemented, covered and documented.

Fixes #166.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 The format of this change log follows the advice given at [Keep a CHANGELOG](https://keepachangelog.com).
 
 ## [Unreleased]
+### Added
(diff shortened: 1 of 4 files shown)


junpataleta (approved):
Thanks, Eloy!

Issue #168, opened 2024-06-17
Expected empty line after the boilerplate check breaks all workplace code


marinaglancy:
Hello! We have a problem with the new moodle-cs
```
FILE: /home/marina/repositories/workplace/wp404-first/moodle/admin/tool/catalogue/lib.php
------------------------------------------------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
------------------------------------------------------------------------------------------------------------------------------------
 16 | ERROR | [x] Boilerplate comment must be followed by a single blank line or end of file
    |       |     (moodle.Files.BoilerplateComment.SingleTrailingNewLine)
------------------------------------------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
------------------------------------------------------------------------------------------------------------------------------------
```

this fails on every single file in workplace because we have a boilerplate that includes workplace clause

![Screenshot from 2024-06-17 09-26-00](https://github.com/moodlehq/moodle-cs/assets/836986/0b4114f9-ebfd-4764-82bb-f8e2dbc94ff0)

our boilerplates look like this and it was perfectly fine before, passed all cs checks. We modified the first line - that this is part of Moodle Workplace and not Moodle and also added our own text in the bottom

stronk7:
Just noting that the problem is not the "customisation" in the first line, that continues being allowed as it was, so far.

Instead, the problem is the "extra" boilerplate text after the official one. And the lack of one blank like after the official part (line 16).

I think that what we should do is to:
1. Allow the "extra" (consecutive) boilerplate to exist, with any contents.
2. Apply the `SingleTrailingNewLine` after the extra boilerplate (instead of current "in the middle" behaviour). 

Ciao :-)

stronk7:
Note that https://github.com/moodlehq/moodle-cs/pull/169 has been created and, hopefully, fixes this.

PR #169, opened 2024-06-17
Allow extra consecutive header lines after the official boilerplate

stronk7:
Once the official boilerplate checks have been performed, allow the "block" to be expanded with any extra information.

As far as the block is a number of consecutive comment lines, exclusively separated by \n, that's allowed.

Note that any spacing different from \n or any non consecutive comment line will stop the sequence.

Fixes: #168


sarjona (approved):
Hi @stronk7!
Thanks for working on this. The patch looks good and it's great you also covered it with unit tests. Well done! :-)

stronk7:
Tata, @sarjona !

PR #171, opened 2024-06-20
Ignore anonymous class docblocks if they have a parent

andrewnicols:
There's usually little need to document an anonymous class when it is part of another method.


stronk7 (approved):
Thanks, Andrew, it looks perfect, hence merging.

PR #173, opened 2024-06-27
Ensure we stop looking for file phpdoc block asap

### LINKED ISSUE #172: Commenting.PackageSniff fix seems broken when first doc comment is an inline one
I'm seeing multiple issues with PackageSniff with code similar to this:

```php
namespace availability_strathenrol;

use enrol_plugin;

class frontend extends \core_availability\frontend {

    protected function get_javascript_init_params($course, \cm_info $cm = null, \section_info $section = null) {
        $instances = enrol_get_instances($course->id, false);

        /** @var enrol_plugin[] $plugins */
        $plugins = enrol_get_plugins(false);
    }
}
```

PHPCS is correctly identifying that there is no package tag:

```
1 | ERROR   | [x] DocBlock missing a @package tag for file frontend.php. Expected @package availability_strathenrol
    |         |     (moodle.Commenting.Package.Missing)
```

But running the fix gives these two issues:

1. It tries to add the tag to the inline docblock where it's clearly not appropriate.

2. In the first pass it adds the tag to the end of the inline comment on the same line:

```php
/** @var enrol_plugin[] $plugins * @package availability_strathenrol
*/
```

This is then detected as still violating the sniff, so it runs a second pass and adds it again on the line with the closing tag.

So the final diff is this:

```diff
-        /** @var enrol_plugin[] $plugins */
+        /** @var enrol_plugin[] $plugins * @package availability_strathenrol
+ * @package availability_strathenrol
+ */
```

stronk7:
Previously, we were allowing the `getDocTagFromOpenTag()` method to advance too much when looking for file phpdoc blocks. Now we stop as soon as any of the stop tokens is found.

Note that this case is very edge one, only reproducible when there are lots of missing class/function phpdoc block, causing some other phpdoc block (type-hinting a variable or whatever) to be picked.

In any case, I think that it's perfectly ok to shortcut the search that way, it can save us some precious iterations.

Fixes #172

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ The format of this change log follows the advice given at [Keep a CHANGELOG](htt
 - The `moodle.NamingConventions.ValidFunctionName` sniff will now ignore errors on methods employing the `#[\Override]` attribute.
 - The `moodle.Commenting.MissingDocblock` sniff no longer warns about missing docs on non-global anonymous classes, for example those written as an instance class in a unit test.
 
+### Fixed
+- Fixed an edge case leading to the file phpdoc block being incorrectly detected by various sniffs.
+
 ## [v3.4.9] - 2024-06-19
 ### Fixed
 - Fixed a recent regression by allowing to the `moodle.Files.BoilerplateComment` sniff to contain "extra" consecutive comment lines immediately after the official boilerplate ends.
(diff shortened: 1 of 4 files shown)


stronk7:
Just amended a few comments and add a new case, via provider, to ensure all cases are covered.

andrewnicols (approved):

PR #175, opened 2024-07-02
Fixes #174. Insert boilerplate correctly where comment on first line.

### LINKED ISSUE #174: BoilerplateCommentSniff causes invalid code if first line has comment
The fix in BoilerplateCommentSniff can cause invalid PHP if there is a comment in the first line.

This code:
```php
<?php // Some comment on first line.

class test {}
```
becomes this:
```php
<?php// This file is part of Moodle - https://moodle.org/
//
// Moodle is free software: you can redistribute it and/or modify
// it under the terms of the GNU General Public License as published by
// the Free Software Foundation, either version 3 of the License, or
// (at your option) any later version.
//
// Moodle is distributed in the hope that it will be useful,
// but WITHOUT ANY WARRANTY; without even the implied warranty of
// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
// GNU General Public License for more details.
//
// You should have received a copy of the GNU General Public License
// along with Moodle.  If not, see <https://www.gnu.org/licenses/>.
// Some comment on first line.

class test {
}
```
Because there is no whitespace after the php tag it is not recognised. Also clearly the boilerplate should be on the line after the opening tag.

micaherne:
This fixes #174 

I've added a test fixture to make sure it works as expected. I'd rather there was a blank line between the boilerplate and the original comment but I can't find a way to do this that doesn't either break other tests, or require a lot of complicated code, so hopefully what it's doing is acceptable.

### DIFF
--- a/moodle/Sniffs/Files/BoilerplateCommentSniff.php
+++ b/moodle/Sniffs/Files/BoilerplateCommentSniff.php
@@ -233,8 +233,16 @@ class BoilerplateCommentSniff implements Sniff
 
     private function insertBoilerplate(File $file, int $stackptr): void
     {
-        $prefix = substr($file->getTokens()[$stackptr]['content'], -1) === "\n" ? '' : "\n";
-        $file->fixer->addContent($stackptr, $prefix . implode("\n", $this->fullComment()) . "\n");
+        $token = $file->getTokens()[$stackptr];
+        $paddedComment = implode("\n", $this->fullComment()) . "\n";
+
+        if ($token['code'] === T_OPEN_TAG) {
+            $replacement = trim($token['content']) . "\n" . $paddedComment;
+            $file->fixer->replaceToken($stackptr, $replacement);
+        } else {
+            $prefix = substr($token['content'], -1) === "\n" ? '' : "\n";
+            $file->fixer->addContent($stackptr, $prefix . $paddedComment);
+        }
     }
 
     private function moveBoilerplate(File $file, int $start, int $target): void
--- a/moodle/Tests/FilesBoilerPlateCommentTest.php
+++ b/moodle/Tests/FilesBoilerPlateCommentTest.php
@@ -212,4 +212,32 @@ class FilesBoilerPlateCommentTest extends MoodleCSBaseTestCase
 
         $this->verifyCsResults();
     }
+
+    public function testMoodleFilesBoilerplateCommentFirstlineComment() {
+        $this->setStandard('moodle');
+        $this->setSniff('moodle.Files.BoilerplateComment');
+        $this->setFixture(__DIR__ . '/fixtures/files/boilerplatecomment/firstline_comment.php');
+
+        $this->setErrors([
+            1 => 'NoBoilerplateComment',
+        ]);
+
+        $this->setWarnings([]);
+
+        $this->verifyCsResults();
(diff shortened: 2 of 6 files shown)


micaherne:
I'm not sure what to do about the code coverage failure here. I haven't really changed the lines it's complaining about, just moved them. They'll presumably only be called now if there is a PHPCS tag on the first line but I'm not sure I quite understand how that is supposed to work so I don't think I can provide a test file.

moodle/Sniffs/Files/BoilerplateCommentSniff.php
```diff
@@ -233,8 +233,16 @@ private function fullComment(): array
...
     {
-        $prefix = substr($file->getTokens()[$stackptr]['content'], -1) === "\n" ? '' : "\n";
-        $file->fixer->addContent($stackptr, $prefix . implode("\n", $this->fullComment()) . "\n");
+        $token = $file->getTokens()[$stackptr];
+        $paddedComment = implode("\n", $this->fullComment()) . "\n";
+
+        if ($token['code'] === T_OPEN_TAG) {
+            $replacement = trim($token['content']) . "\n" . $paddedComment;
+            $file->fixer->replaceToken($stackptr, $replacement);
+        } else {
```
stronk7, line 242: Hi @micaherne ,

I've just seen that the new code now leaves these 2 few lines uncovered, when they were previously covered? Can you ensure that the uncovered code can happen and, because of that, it should have some fixture covering it?

Ciao :-)

stronk7 (changes requested):

moodle/Sniffs/Files/BoilerplateCommentSniff.php
stronk7: Or, with other words... why the existing fixtures (before this patch), don't cover those lines when they previously were? I mean, is "smells" that they have become uncovered with your patch.

micaherne: Thanks for the feedback @stronk7! I wasn't sure what to do about the code coverage but I added a test file with a phpcs tag and it's causing that branch to run.

moodle/Tests/FilesBoilerPlateCommentTest.php
```diff
@@ -212,4 +212,32 @@ public function testMoodleFilesBoilerplateCommentTrailingWhitespaceMissing() {
...
+    }
+
+    public function testMoodleFilesBoilerplateCommentWithPhpcsTag() {
+        $this->setStandard('moodle');
+        $this->setSniff('moodle.Files.BoilerplateComment');
+        $this->setFixture(__DIR__ . '/fixtures/files/boilerplatecomment/with_phpcs_tag.php');
+
+        $this->setErrors([
+                             1 => 'NoBoilerplateComment',
+                         ]);
```
stronk7, line 237: Please, amend this indentation.

stronk7 (approved):
Thanks @micaherne , I think that's all we needed, now we have a "proof" of the use case behind that code branch.

I'm approving this now, would be great if the remaining indentation detail is fixed. Will merge then, thanks!

micaherne:
Oops I don't know how I managed that! That's it fixed now :)

PR #177, opened 2024-07-04
Fixes #176. Match blank boilerplate lines correctly.

### LINKED ISSUE #176: Inserted boilerplate gets broken if there is a CRLF line break after the opening tag
I'm seeing an issue with the fix for the Files.BoilerplateComment sniff where there is a Windows line ending after the opening `<?php` tag.

Running phpcbf with the moodle standard on this code:

```php
<?php
```
gives this:
```php
<?php
// This file is part of Moodle - https://moodle.org/
// Moodle is free software: you can redistribute it and/or modify
// Moodle is free software: you can redistribute it and/or modify
// it under the terms of the GNU General Public License as published by
// the Free Software Foundation, either version 3 of the License, or
// (at your option) any later version.
// but WITHOUT ANY WARRANTY; without even the implied warranty of
// Moodle is distributed in the hope that it will be useful,
// but WITHOUT ANY WARRANTY; without even the implied warranty of
// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
// GNU General Public License for more details.
//
// You should have received a copy of the GNU General Public License
// along with Moodle.  If not, see <https://www.gnu.org/licenses/>.
```

This doesn't happen if running the boilerplate sniff only (`vendor/bin/phpcbf --standard=moodle --sniffs=moodle.Files.BoilerplateComment`) and from looking at the processing output it looks as if it may be an interaction with moodle.Commenting.InlineComment which seems to remove the blank lines from the boilerplate, which BoilerplateComment then adds back in wrongly.

I think there may be a couple of buggy things going on here so I'll try to work out in more detail what is happening.

micaherne:
This patch fixes an issue that contributes to #176 where lines of the boilerplate that should be blank are not identified as wrong if they have text in them. 

This is because we are only checking that the lines start with the expected text. I'm not sure why this is but it was this way in the original code before I added the fixes.

I think this fixes #176 in a way, although there is still a weird chain of fixing going on when there's a CRLF in the opening PHP tag. I'll add some more to that issue about it just in case it comes up again.

### DIFF
--- a/moodle/Sniffs/Files/BoilerplateCommentSniff.php
+++ b/moodle/Sniffs/Files/BoilerplateCommentSniff.php
@@ -288,6 +288,11 @@ class BoilerplateCommentSniff implements Sniff
      */
     private function regexForLine(string $line): string
     {
+        // We need to match the blank lines in their entirety.
+        if ($line === '//') {
+            return '/^\/\/$/';
+        }
+
         return str_replace(
             ['Moodle', 'https\\:'],
             ['.*', 'https?\\:'],
--- a/moodle/Tests/FilesBoilerPlateCommentTest.php
+++ b/moodle/Tests/FilesBoilerPlateCommentTest.php
@@ -240,4 +240,26 @@ class FilesBoilerPlateCommentTest extends MoodleCSBaseTestCase
 
         $this->verifyCsResults();
     }
+
+    public function testMoodleFilesBoilerplateCommentMissingLines() {
+        $this->setStandard('moodle');
+        $this->setSniff('moodle.Files.BoilerplateComment');
+        $this->setFixture(__DIR__ . '/fixtures/files/boilerplatecomment/missing_lines.php');
+
+        $this->setErrors([
+            3 => 'moodle.Files.BoilerplateComment.WrongLine',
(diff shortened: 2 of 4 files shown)


stronk7:
Hi @micaherne , regarding CRLF ended lines... I think we shouldn't apply for any specific workaround or fix ever.

Moodle files are supposed to be, always, LF ended (Unix ended) and I think that the correct and safe thing to do is to, always, assume that's the case.

If there is any file having other line endings... then for sure some other sniff (`Generic.Files.LineEndings` is part of our standard) should detect that and report it, but we cannot make all the code conditional on those wrong line endings.

So it's good that we aren't adding any specific CRLF code here (or elsewhere, ever). IMO. Because that code is not allowed. As simple as that.

Ciao :-)

stronk7 (approved):
Other than the comment above, that is just a reflexion... the changes look spot-on and are perfectly covered, so I think this is ready to be merged.

Thanks!

micaherne:
Thanks @stronk7! I totally agree about CRLF endings - I don't think it makes any sense to add extra code to work around them - I was just a bit confused about how they were affecting the other fixes. But anyway it seems to be giving the expected output with them anyway so I'm happy with that 😄

PR #179, opened 2024-07-22
Bump to phpcs 3.10.2

stronk7:
I've also reviewed other dependencies and only PHPCompatibility is slightly "old" (May 2024), so decided not to bump it.


andrewnicols:
My many apologies Eloy. I'm closing this as it has now been superceded by #203.

PR #182, opened 2024-08-29
ValidVariableNameSniff: freshen allowed globals

### LINKED ISSUE #181: Allowed globals is incomplete?
In short, the $FULLSCRIPT global is missing from the list of allowed variable names.

As of August 22, 2024, here are the allowed globals:
https://github.com/moodlehq/moodle-cs/blob/89ff0acd727f8611cbafe5aea524614090b24ad9/moodle/Sniffs/NamingConventions/ValidVariableNameSniff.php#L32-L37

While I see the PHP superglobals ($_GET, $_POST, etc.) and the main Moodle global objects ($CFG, $DB, $USER, etc.) there are a few Moodle global literals like $ME, $FULLME, $SCRIPT that are present.

Today was my first day seeing those literals, because I was looking for origin of $FULLSCRIPT. I don't know the difference between the four and I don't see any uses in core of $FULLSCRIPT outside of it being defined and initialized. Someone else didn't know, so they incorrectly lowercased the variable thinking that it was a plugin-specific variable, which may have caused some issues.

jrchamp:
- Group by global type with comment.
 - Add comment to removed globals.
 - Add missing FULLSCRIPT literal.
 - Add missing XMLDB object.

Fixes #181

I didn't remove any that were already present, but I have opinions. 😈


andrewnicols (approved):
Thanks for that change - much appreciated.

Issue #58, opened 2023-09-15
Raise Generic.Arrays.DisallowLongArraySyntax to error - September 2024


stronk7:
In #34 we started checking for PHP long array syntax (September 2023).

As we have done with other new checks, it was created:

- As warning within the `moodle` standard.
- As error within the `moodle-extra` standard.

This issue is about to make it an error in both standards after 1 year of "grace" period.

Ciao :-)

PR #201, opened 2025-07-02
Apply PSR-12 rules per MDLSITE-7597

andrewnicols:
Most of the PSR-12 rules are now standard for Moodle with the following exceptions:

- line length
- variable naming
- class naming
- opening brace on new line
- side effects
- use of `else if` instead of `elseif`

Additionally we currently exclude a small number of PSR-12 rules, including:

- constant visibility may be ignored on class members.


moodle-extra/ruleset.xml
```diff
@@ -99,4 +38,10 @@
         </properties>
     </rule>
 
+    <!-- Moodle contains a lot of code which pre-dates PHP 7.1 and did not support constant visibility -->
+    <rule name="PSR12.Properties.ConstantVisibility"/>
+
+    <rule ref="PSR12.Properties.ConstantVisibility.NotFound">
+        <severity>5</severity>
+    </rule>
```
jrchamp, line 46: This is a duplicate of lines 12-14. I believe only one of these is needed.

moodle-extra/ruleset.xml
```diff
@@ -99,4 +38,10 @@
         </properties>
     </rule>
 
+    <!-- Moodle contains a lot of code which pre-dates PHP 7.1 and did not support constant visibility -->
+    <rule name="PSR12.Properties.ConstantVisibility"/>
```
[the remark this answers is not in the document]
andrewnicols: Removed as this is a dupe.

moodle/ruleset.xml
```diff
@@ -13,6 +13,16 @@
         <type>warning</type>
```
jrchamp, line 13: If the rule type=error is being removed from moodle-extra to resolve #58, then this should be an error in moodle.

Otherwise, it probably shouldn't be removed from moodle-extra.

PR #204, opened 2025-07-29
Add missing valid tags for @var blocks

andrewnicols:
It is perfectly acceptable to:

- provide a @link to elsewhere;
- declare when a var was made available with the @since tag
- declare that a var was @deprecated


paulholden (approved):
Cheers Andrew :+1:

PR #218, opened 2025-11-06
Allow use of the $GLOBALS superglobal

cameron1729:


andrewnicols (approved):

jrchamp:
Use of global variables in general is something I typically discourage these days. There's not much discussion here, so I'm guessing it happened in another forum.

My question: Is $GLOBALS being added here for backward compatibility with existing code? Or are we saying that $GLOBALS is unconditionally allowed by Moodle Coding standards?

Personally, I would recommend temporarily allowing, but discouraging the use of $GLOBALS because access to these values is already available from the `global` keyword and which serves as a very clear semantic highlight of external dependencies for a method. Then, over years, usage could be phased out and changes to global state could be less problematic.

andrewnicols:
Hi @jrchamp,

Yes, we do discourage its use, but it is still in use in a few places and automatically fixing it will break things, so we to allow it.

jrchamp:
Perfect! Thank you for confirming.

cameron1729:
Worth noting: the sniff says nothing about whether you should use `$GLOBALS` or not. PHPCS simply reports that `$GLOBALS` should be lower case - which if course it shouldn't be; and if you followed the advice and changed it, your code would break.

jrchamp:
@cameron1729 Excellent point! 🧠 In that sense, this sniff should include all of the special variables from both PHP itself and Moodle core. I can't begin to thank you for highlighting that note. ❤️ 

As far as discouraging use, I agree that seems that would be a different Sniff. Perhaps something new like `moodle.PHP.DeprecatedVariables`. Either way, that's probably something that should be its own issue.

cameron1729:
No worries @jrchamp - happy to help :)

I think a new sniff makes sense too. Maybe part of the moodle-extra coding style.

PR #223, opened 2025-12-28
Improve PHPDoc Types

james-cnz:
I've run a basic PHPDoc type checker over the code, and tidied up a few things.  I think it's probably good for moodle-cs to set a good example.

Note:  I know the PHPCS Sniff function process() specifies its return type as void|int, but AFAIK, void is a stand-alone type, and int|null is the correct type.

### DIFF
--- a/moodle/Sniffs/Commenting/CategorySniff.php
+++ b/moodle/Sniffs/Commenting/CategorySniff.php
@@ -32,6 +32,8 @@ class CategorySniff implements Sniff
 {
     /**
      * Register for open tag (only process once per file).
+     *
+     * @return (int|string)[]
      */
     public function register() {
         return [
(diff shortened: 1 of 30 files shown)


jrchamp:
None of the process() methods that I looked at had a `return` that was anything other than `void` (i.e. `return;`). Where are you seeing int returns? Also, if nullable int were the correct return value, I'd lean towards `?int` over `int|null` because - to me - it implies "sometimes you get an integer" as opposed to "sometimes you get an integer which means A and sometimes you get null which means B".

I hope that helps! I'm sure others will have more comprehensive feedback.

james-cnz:
Yeah, I don't think any of the process methods in moodle-cs return int.  It would also be valid to specify the return type as just null.  It's in the base method, though, and I didn't think there was any point narrowing it.
https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Sniffs/Sniff.php
The return statements should probably say return null; for good style, but PHP defaults to null if trying to read the return value of a function when none was provided.
I used int|null because it conforms to PSR-5, which I've been told the Moodle code base uses, so I've leaned into this style where possible.  Maybe it's not appropriate for moodle-cs, though.

PR #225, opened 2026-03-12
Fix imported Override attribute handling

cameron1729:
Methods marked with `#[Override]` were still triggering some sniffs when the attribute was imported with use `Override;` because the checks were looking for exactly `\Override` and missed imported usages like `#[Override]`.

This patch updates the sniff to use the qualified attribute name instead of the raw token.


jrchamp:
I love that this patch converts multiple instances of code duplication to use the utility helper method that is designed for the task. ❤️

PR #227, opened 2026-06-19
Port MoodlePHPdoc check function types check

james-cnz:
This is a PHPDoc types check intended to be similar in scope to the MoodlePHPdoc check one.

It is a bit more permissive in a few respects:
- It makes some basic effort to allow PHPDoc types to be subtypes of the native type.  (I believe this is standard to PHPDoc types.)
- It allows mixed type arrays, e.g. (a|b)[] .  (This is supported by PHP-FIG PSR-5.)
- It allows PHPDoc param annotations to contain pass by reference or splat.

But, AFAIK, it is only stricter in one way:
- If the native type contains null, the PHPDoc type must too.  [EDIT: Sorry, I mean the other way round.  If the Doc type contains null, the native type must too.]
[EDIT:  It looks like the existing checks fail to check PHPDocs that are separated from the function by an attribute, so it's also stricter in that way.]

As with the existing test, it makes no effort to syntax check types, check namespaces, match return types, or do any checking on variables.
And it passes existing tests.
So I don't think it should cause too much disruption.


james-cnz:
I forgot about parentheses in types.

james-cnz:
I've made the type simplification aware of well-formedness of types.  It doesn't make much difference now, but would make it easier to include a check that types are well-formed in the future, if desired.

james-cnz:
I forgot about the type "$this".

james-cnz:
I've improved subtype recognition a bit, to support, e.g. MDL-83375.

james-cnz:
Improved error messages a bit, to help address MDLSITE-3694.

PR #234, opened 2026-09-14
update PHP requirements

ziegenberg:
The current Moodle LTS has a PHP min requirement of 8.1, so I think it's okay to drop support for PHP 7.4. Plus, we should test with PHP 8.4 and PHP 8.5 now.


lameze (commented):
Thanks Daniel, this patch looks spot on.

